PR #76, opened 2022-05-18
Update check-dist.yml

Aib666:


febuiles:
Closing since the Action is already being used in a different file (`dependency-review.yml`).

PR #106, opened 2022-06-07
Adding allow and deny lists

febuiles:
### Introduction
This PR introduces two new configuration options, `allow_licenses` and `deny_licenses`. They allow the user to specify lists of allowed and denied licenses for their project. Here's an example config file:

```yaml
fail_on_severity: low
allow_licenses:
  - 'GPL 3.0'
  - 'BSD 3 Clause'
  - 'MIT'
```

In addition to the severity check, we will now block the PR if any of our dependencies has a license that's different to the ones specified in the `allow_licenses` list. 

`deny_licenses` is the blocklist equivalent, a PR will be blocked if any of the dependencies introduce a license that's part of the `deny_licenses` list:

```yaml
fail_on_severity: critical
deny_licenses:
  - "LGPL 2.0"
  - "BSD 2 Clause"
```

These two options are mutually exclusive, users need to pick the one that suits them better. An error will be raised if both options are specified in the configuration file.

### Config file considerations

1. The PR contains no public documentation (README updates). These will be added once we do a release.
2. We will be moving away from having a separate configuration file for this Action, so please consider this separate format to be temporary. Going forward we'll be using lists of licenses separated by commas directly in the workflow file
```yaml
- name: Dependency Review
  uses: actions/dependency-review-action@main
  with:
    allowed_licenses: GPL 2, BSD, MIT
```

### Unknown licenses
There's a good number of dependencies where we don't know the license of the package. The work in this PR discards these dependencies from the final list of results. 

In a follow-up PR I'll add more comprehensive support for unknown licenses: We will always show a message mentioning the number of dependencies with unknown licenses. Users can also add `unknown` to their `deny_list` and this will block PRs when dependencies with unknown licenses are introduced:

```yaml
- name: Dependency Review
  uses: actions/dependency-review-action@main
  with:
    deny_licenses: GPL 2, unknown
```


cnagadya:
Awesome :) ! This is such a great addition to the action. Checking the licences was such a big deal on my previous team

src/licenses.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+    if (allowLicenses.length > 0) {
+      if (!allowLicenses.includes(license)) {
+        disallowed.push(change)
+      }
+    } else if (failLicenses.length > 0) {
+      if (failLicenses.includes(license)) {
+        disallowed.push(change)
+      }
+    }
```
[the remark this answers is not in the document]
febuiles: > So, my reading of the logic here is that empty allowLicenses and denyLicenses lists are ignored as filters?

Correct. I'm hesitant to use null because the following is not obvious to me:

> `allowLicenses = []` would mean the obvious: the list of allowed licenses is empty, aka no licenses are allowed.

I was using `null` originally, but I had to insert a few checks to satisfy the compiler and ended up with a version longer than what I have above. 

I don't have strong feelings regarding this, I can change it either here or in a follow-up PR.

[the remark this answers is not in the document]
hmaurer: @febuiles regarding `allowLicences = []`, I think what it boils down to is I don't think `[]` should have a distinctly different meaning to a non-empty array. Consider the following:

```
allow_licenses: []
```

The most natural interpretation of the above, in my view, is "no licenses are allowed". But under the implementation you are proposing, this would behave as "all licenses are allowed". As a user, I have to know that the action is treating `[]` as synonymous with the property not being specified to make sense of the behaviour. I'd rather avoid the potential confusion.

src/main.ts
```diff
@@ -99,4 +116,27 @@ function renderSeverity(
   return `${styles.color[color].open}(${severity} severity)${styles.color[color].close}`
 }
 
+function printLicensesError(
+  changes: Array<Change>,
+  allowLicenses: Array<string> | undefined,
+  denyLicenses: Array<string> | undefined
+): void {
```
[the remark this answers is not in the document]
febuiles: Fixed in the same manner as the other function.

src/licenses.ts
```diff
@@ -0,0 +1,39 @@
...
+    let license = change.license
+    // TODO: be loud about unknown licenses
+    if (license === null) {
+      continue
+    }
+    if (allow !== null) {
+      if (!allow.includes(license)) {
+        disallowed.push(change)
+      }
+    } else if (deny !== null) {
```
lseppala, line 31: Following the logic, this means that the allow list always super-cedes deny lists. That is, if there's any allow list, the deny list won't be used (the `if-else` ensures that the else branch will not be taken if allow !== null).

Is this the intended behavior? Fine if it is, but let's be sure to make that behavior explicit in documentation and tests.

src/licenses.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  let disallowed: Change[] = []
+
+  for (const change of changes) {
+    let license = change.license
+    // TODO: be loud about unknown licenses
+    if (license === null) {
+      continue
+    }
+    if (allow !== null) {
```
lseppala, line 27: This condition will never *not* be true. 

For optional object parameters like `allow?: Array<string>`, the value will be an array of string (if defined) or `undefined` (if not defined).

In JS, `undefined` is not strictly equal to `null` (`undefined !== null` is true). However, `undefined` is loosely equal to `null` (`undefined != null` is false). 

Beware that an empty array is never equal to null, either with strict or loose equality. So passing an empty array would also evaluate this to true.

Might be worth a few more unit tests to make sure this is correct.

src/licenses.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param { { allow?: string[], deny?: string[]}} licenses An object with `allow`/`deny` keys, each containing a list of licenses.
+ * @returns {Array<Change} The list of denied changes.
+ */
+export function getDeniedLicenseChanges(
+  changes: Array<Change>,
+  licenses: {
+    allow?: Array<string>
+    deny?: Array<string>
+  }
+): Array<Change> {
```
lseppala, line 16: I really appreciate programs that correct _all_ of the errors and display them all at once. It's really frustrating to get an error, fix it, run the program again, get another error, and repeat. I'm glad you're doing this 😄

lseppala (changes requested):
There's a bit of a flaw in your `getDeniedLicenseChanges` logic, but otherwise, looks like good work!

src/licenses.ts
febuiles: This is on purpose, from the PR description: 

> These two options are mutually exclusive, users need to pick the one that suits them better. An error will be raised if both options are specified in the configuration file.

Zod takes care of [doing the actual validation](https://github.com/actions/dependency-review-action/blob/adding-lists/src/schemas.ts#L40-L43). [Here's a test](https://github.com/actions/dependency-review-action/blob/adding-lists/__tests__/config.test.ts#L27-L31) that covers this use case from a higher level. 

I've added more information in the function's comment in 6b5518a.

src/licenses.ts
febuiles: The original intent was to know if we were dealing with default parameters or if the user was actually providing a list. @hmaurer called out this point yesterday too, and I made some changes in cc22dcd. What do you think? 

[These two tests](https://github.com/actions/dependency-review-action/blob/adding-lists/__tests__/licenses.test.ts#L49-L59) now exercise this code path.

src/main.ts
```diff
@@ -99,4 +113,35 @@ function renderSeverity(
...
+
+  if (allow.length > 0) {
+    core.info('\nAllowed licenses: ' + allow.join(', ') + '\n')
+  }
+
+  if (deny.length > 0) {
+    core.info('\nDenied licenses: ' + deny.join(', ') + '\n')
+  }
+
+  core.info('The following dependencies have incompatible licenses:\n')
```
elireisman, line 139: same

src/main.ts
```diff
@@ -99,4 +113,35 @@ function renderSeverity(
...
+  if (changes.length == 0) {
+    return
+  }
+
+  let {allow = [], deny = []} = licenses
+
+  core.info('Dependency review detected incompatible licenses.')
+
+  if (allow.length > 0) {
+    core.info('\nAllowed licenses: ' + allow.join(', ') + '\n')
```
elireisman, line 132: Optional: I wonder if it might be least confusing to just display the ineligible licenses the scan found?

__tests__/licenses.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  ]
+}
+
+let rubyChange: Change = {
+  change_type: 'added',
+  manifest: 'Gemfile.lock',
+  ecosystem: 'rubygems',
+  name: 'actionsomething',
+  version: '3.2.0',
+  package_url: 'somerubypurl',
```
elireisman, line 30: 🤔 total nit/optional: since this Action is open-source, it might be helpful to use a real(istic) PURL and similar for other fields, in order for these tests to act a bit more as documentation about the API data for end users (and possibly future contributors)

src/licenses.ts
lseppala: Perfect, thanks for clarifying 👍

src/licenses.ts
lseppala: I think this addresses the major concern. Passing an empty list will still be true for `allow !== undefined`, which may lead to surprising behaviors if someone were to call this function with an empty list, but a lesser concern.

src/main.ts
febuiles: Fixed in 571f236.

src/main.ts
```diff
@@ -99,4 +113,35 @@ function renderSeverity(
...
+    deny?: Array<string>
+  }
+): void {
+  if (changes.length == 0) {
+    return
+  }
+
+  let {allow = [], deny = []} = licenses
+
+  core.info('Dependency review detected incompatible licenses.')
```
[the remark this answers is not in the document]
febuiles: Fixed in 571f236, thank you!

src/main.ts
febuiles: I think good to show the users what's in their config files in case there's a failure, saves you a trip to the YAML file. If this causes any type of confusion I'll be happy to change this behavior!

__tests__/licenses.test.ts
febuiles: I like this idea, fixed in eebebcd 👍

PR #112, opened 2022-06-14
Move configuration file location

febuiles:


src/config.ts
```diff
@@ -1,29 +1,29 @@
...
+  let options: ConfigurationOptions = {}
 
-export function readConfigFile(
-  filePath: string = CONFIG_FILEPATH
-): ConfigurationOptions {
   // By default we want to fail on all severities and allow all licenses.
-  const defaultOptions: ConfigurationOptions = {
-    fail_on_severity: 'low'
-  }
+  const severity = core.getInput('fail-on-severity') || 'low'
```
febuiles, line 12: I don't like having to do the OR here, and I'd expect Zod to take care of this, but removing the clause returns `undefined`.

elireisman (approved):

PR #141, opened 2022-07-04
Use fixed major version for node types

kachick:
This action works under node 16, is types only updating to node 18 intentional? 👀  https://github.com/actions/dependency-review-action/blob/26b790870174b01a5110cc2be9ad84ec2374e5b5/action.yml#L20
https://github.com/actions/dependency-review-action/blob/26b790870174b01a5110cc2be9ad84ec2374e5b5/.github/workflows/check-dist.yml#L26-L29

ref: https://github.com/actions/dependency-review-action/pull/137

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,6 @@ updates:
     directory: /
     schedule:
       interval: daily
+    ignore:
+      - dependency-name: '@types/*'
+        update-types: ['version-update:semver-major']
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,5 +10,5 @@ updates:
     schedule:
       interval: daily
     ignore:
-      - dependency-name: '@types/*'
+      - dependency-name: '@types/node'
         update-types: ['version-update:semver-major']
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -9,3 +9,6 @@ updates:
     directory: /
     schedule:
       interval: daily
+    ignore:
+      - dependency-name: '@types/*'
```
febuiles, line 13: Should we be explicit about the name (e.g.)

```suggestion
      - dependency-name: '@types/node'
```

or is there a reason to use the wildcard here?

febuiles (commented):
Left a question (I'm new to Node), but thanks for catching this!

.github/dependabot.yml
kachick: To prevent making unexpected different major versions between implementation and DefinitelyTyped. Current `package.json` does not depend on another DefinitelyTyped. However https://github.com/actions/dependency-review-action/pull/131 is integrating a new one.
As far as I know, this wildcard is useful when you only want minor and patch PRs from dependabot . Of course I can update this to focus only on the `node` issue.

PR #142, opened 2022-07-04
Add CI workflow and fix lint errors

kachick:
This PR has been extracted from https://github.com/actions/dependency-review-action/pull/131#issuecomment-1166172866 to fix current errors.

PR #146, opened 2022-07-05
Enable prettier and recommend eslint in vscode workspace config

kachick:
I think this helps to use already integrated tools🙏

https://github.com/actions/dependency-review-action/blob/26b790870174b01a5110cc2be9ad84ec2374e5b5/package.json#L9-L11

ref: https://github.com/actions/dependency-review-action/pull/142


kachick:
Thank you!

PR #181, opened 2022-08-04
Show vulnerabities and license information on the job summary.

tspascoal:
Users can see the results that were found directly on the job summary

All the results are grouped by manifest.

It shows a table with vulnerable packages, together with package version,
the vulnerabily info and it's severity.

Shows info about package licenses, which packages have a non allowed license,
and the list of packages with unknown licenses.


febuiles:
@tspascoal Thanks! Do you mind adding a screenshot of how this would look like, or linking to an Action run we can use to see the behavior before merging?

src/summary.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export function addSummaryToSummary(
+  addedPackages: Changes,
+  licenseErrors: Change[],
+  unknownLicenses: Change[]
+): void {
+  core.summary
+    .addHeading('Dependency Review')
+    .addRaw(
+      `We found ${addedPackages.length} vulnerable package(s), ${licenseErrors.length} package()s with incompatible licenses, and ${unknownLicenses.length} package(s) with unknown licenses.`
```
ericcornelissen, line 13: Looks like there's a typo in "package()s":

```suggestion
      `We found ${addedPackages.length} vulnerable package(s), ${licenseErrors.length} package(s) with incompatible licenses, and ${unknownLicenses.length} package(s) with unknown licenses.`
```

tspascoal: Good catch.

Thank you

PR #165, opened 2022-07-21
Support user-provided base/head refs & non-PR workflows

WillDaSilva:
This PR implements @febuiles' suggestion from https://github.com/actions/dependency-review-action/issues/30#issuecomment-1147371846:

> Maybe it'd make sense to add extra optional parameters to the Action so it accepts head/base revisions. Leaving this open while considering other options (PRs to mitigate this issue in the meantime are welcome!)

Closes https://github.com/actions/dependency-review-action/issues/30

PR #200, opened 2022-08-18
Support user-provided base/head refs & non-PR workflows 

febuiles:
@WillDaSilva apologies, I closed #165 by mistakenly changing the merge HEAD to this branch in order to be able to push some minor changes.

Do you have a sample workflow file we could use to showcase the behavior in the README and Release Notes?


WillDaSilva:
> Do you have a sample workflow file we could use to showcase the behavior in the README and Release Notes?

@febuiles Not exactly. At Meltano we've been using my fork of this action for 25 days now, and haven't run into any problems: https://github.com/meltano/meltano/blame/6de7c109b89dd1c0e7ace993d2934a51c4fb7867/.github/workflows/dependency-review.yml#L28

We only needed it to run the action for `pull_request_target` events though, so we aren't testing out setting a custom `head-ref` and `base-ref`. We could do so to support running that step of the workflow on `workflow_dispatch` events, but that might be a bit confusing since the following step of the workflow would still only run for the checked out ref.

PR #228, opened 2022-09-06
Add external configuration file

febuiles:
We are bringing back the behavior from https://github.com/actions/dependency-review-action/pull/91 where users can customize the action through an external configuration file. 

### Acceptance Criteria
- [x] Users can supply a custom path to an external configuration file.
- [x] If an external file is specified, the options in there take precedence over the inline ones.


src/config.ts
```diff
@@ -16,14 +24,29 @@ function parseList(list: string | undefined): string[] | undefined {
...
-    .parse(getOptionalInput('fail-on-severity'))
+  const externalConfig = getOptionalInput('config-file')
+  if (externalConfig !== undefined) {
+    const config = readConfigFile(externalConfig)
+    // the reasoning behind reading the inline config when an external
+    // config file is provided is that we still want to allow users to
+    // pass inline options in the presence of an external config file.
+    const inlineConfig = readInlineConfig()
+    // the external config takes precedence
+    return Object.assign({}, inlineConfig, config)
```
sarahkemi, line 35: TIL about `Object.assign`, nicee.

sarahkemi (approved):
🙌🏽 looks great, this is going to be so useful!!

src/config.ts
febuiles: TBH I brought this from plain JS and I'm sure there's a prettier way to do this in 2022 with TS (@brphelps  @hmaurer maybe?)

PR #243, opened 2022-09-19
Filter blocking dependency changes by scopes

sarahkemi:
The dependency review API now returns a dependency's `scope` in the endpoint response! 

```json
{
    "change_type": "added",
    "manifest": "pom.xml",
    "ecosystem": "maven",
    "name": "org.postgresql:postgresql",
    "version": "42.3.2",
    "package_url": "pkg:maven/org.postgresql/postgresql@42.3.2",
    "license": "BSD-2-Clause",
    "source_repository_url": "https://github.com/pgjdbc/pgjdbc",
    "scope": "runtime",
}
```

Let's utilize it to allow users of the action to filter out which types of dependencies they actually want to block on. For example, users will now be able to decide that they don't want to block a PR on a vulnerability if it's on a dependency only utilized in development.

To create this functionality I added a new `fail-on-scopes` configuration input that can take a list of the 3 possible scope types (`unknown`, `runtime`, and `development`). We then parse the list and run the dependency changes from the API through `filterChangesByScopes`that will return only changes that that are in the specified scopes. We then run through the standard filtering of severity and license validation.

To ensure that we don't break the action for GHES users who won't have access to the current REST API version with `scope` just yet, we treat all their dependency changes as `runtime` and detect vulns/licenses on them. This emulates the preexisting behavior.


febuiles (approved):
This looks great @sarahkemi! I've tested it with both runtime and runtime/development vulnerabilities ([PR here](https://github.com/future-funk/bug-free-sniffle/pull/3)) and everything works as expected.

Development Only:
Screenshot 2022-09-20 at 12 43 46

Development/Runtime:
Screenshot 2022-09-20 at 12 46 02

PR #248, opened 2022-09-22
Add scanned deps

### LINKED ISSUE #183: Log/document which sources are being analyzed/supported
After adding this action to one of my project I was curious what it was scanning. Unfortunately I was not able to figure out what is actually being scanned (other than [the image from the README](https://user-images.githubusercontent.com/2161/161042286-b22d7dd3-13cb-458d-8744-ce70ed9bf562.png), suggesting it's scanning `package.json`; and a log about a GitHub Actions workflow, suggesting it's scanning GitHub Actions workflow files). I tried enabling debug logging, but the logs still didn't tell me what was being analyzed.

In that light, I have two suggestions:

1. Document which sources **can** be analyzed in the README (or similar).
1. Log which sources **are** being analyzed during a run of the Action.

febuiles:
This PR introduces some features to display more information about what's been scanned in a run. It closes #183. 

It introduces a new section for scanned manifests and dependencies that list *all* the files and dependencies that change and were scanned, regardless of their licenses or vulnerabilities.

Screenshot 2022-09-27 at 12 27 41

And also an equivalent version (which could use more work) on the summary page:
Screenshot 2022-09-27 at 12 28 16


src/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+): Map<string, Changes> {
+  const dependencies: Map<string, Changes> = new Map()
+  for (const change of changes) {
+    const manifestName = change.manifest
+
+    if (dependencies.get(manifestName) === undefined) {
+      dependencies.set(manifestName, [])
+    }
+
+    dependencies.get(manifestName)?.push(change)
```
febuiles, line 14: The type system is incapable of recognizing that this has already been set above, hence the questionable `?`

src/summary.ts
```diff
@@ -150,14 +151,21 @@ export function addLicensesToSummary(
...
-    return text
+  const summary = core.summary
+    .addHeading('Scanned Dependencies')
+    .addRaw(`We scanned ${dependencies.size} manifest files:`)
+
+  for (const manifest of manifests) {
+    const deps = dependencies.get(manifest)
+    if (deps) {
+      const dependencyNames = deps.map(
+        dependency => `<li>${dependency.name}@${dependency.version}</li>`
```
febuiles, line 166: Don't love creating these by hands, but as far as I can tell the [Actions Summary JS API](https://github.com/actions/toolkit/blob/main/packages/core/src/summary.ts) is not publicly documented (?), and I couldn't find a way in there to create nested lists inside a loop.

febuiles:
Merging in order to start work in license improvement, not doing a new release just yet in case there's any open feedback.

Issue #133, opened 2022-06-27
Supporting whitelisting CVE's / Dependencies


AtzeDeVries:
It would be nice to be able to ignore / whitelist certain dependencies. This way we can keep working if no fix exist yet/

AtzeDeVries:
I thought about having some config which sits with the code (so not in .github dir) which contains a whitelist (or ignorelist) which at least contains

* CVE
Possible also
* End date until whitelist is valid (so required a review after x months)

And also i don't now how the CVE's on dependencies work, but i guess CVE's on a dependency can also be updated (added text, change of severity) so it would be nice to link the ignore  a certain version of the CVE.

febuiles:
We'll likely move our config options (licenses and allow/denylist) to an external config file in the upcoming months to make this Action easier to install in big organizations. Will take a look at this when we do the migration.

AtzeDeVries:
Thnx for taking a look at it!

LiuVII:
I don't see how this can be fully utilized without whitelisting really, looks like a must for anything beyond the MVP phase.
Is there some workaround in a meantime at least?

febuiles:
@LiuVII what do you think about adding a new config option (a list called `ignore`?) that is checked against the `advisory_ghsa_id` field of the [API response](https://github.com/actions/dependency-review-action/blob/main/src/schemas.ts#L18)? We don't receive CVEs in the API response, hence the fallback to GHSA ids.

If you need help getting a PR started please let me know!

PR #251, opened 2022-09-22
Filter by vulnerability allow-list

sarahkemi:
This should close https://github.com/actions/dependency-review-action/issues/133.

This PR introduces a new `allow-ghsas` config input that takes a list of GHSA ids and ignores changes if they only have vulnerabilities with advisory ids from the list. It follows the typical filter pattern of having a new filter function (`filterOutAllowedAdvisories`) that is utilized in the main flow to whittle down the list of desired changes to verify and possibly fail on.


README.md
```diff
@@ -130,12 +132,29 @@ https://docs.github.com/en/rest/licenses.
...
+
 ```yaml
 deny-licenses:
   - LGPL-2.0
   - BSD-2-Clause
 ```
 
+### allow-ghsas
+
+Add a custom list of GHSA ids you do not want the action to block on.
```
febuiles, line 144:
```suggestion
Add a custom list of GitHub Advisory IDs that can be skipped during detection. 
```

febuiles: Tiny nit: the Action does not block anything, it just reports a failure. Enforced checks (repo admin feature I think) is what blocks the actual PR.

febuiles (approved):
:shipit:!

PR #279, opened 2022-10-11
Minor Action Fixes

cnagadya:
Minor changes to:

-  return original changes list when no [allow-ghsas](https://github.com/actions/dependency-review-action#allow-ghsas) value is set in config file
- correctly display `dependency name` instead of `manifest name`


cnagadya:
@febuiles wins this round with #280

PR #294, opened 2022-10-17
Add support for SPDX expressions

cnagadya:
Integrates:
- [spdx-expression-parse](https://www.npmjs.com/package/spdx-expression-parse) to ensure licenses added in the allow-licenses and deny-licenses are [valid SPDX license expressions](https://spdx.dev/spdx-specification-21-web-version/#h.jxpfx0ykyb60)
-  [spdx-satisfies ](https://www.npmjs.com/package/spdx-satisfies) to make license comparisons more robust. However, there is [an existing bug ](https://github.com/jslicense/spdx-satisfies.js/issues/14)in the package causing [unexpected results](https://github.com/actions/dependency-review-action/blob/57404e86cb7caff2641008eeca0e0df22173b78f/src/licenses.ts#L62-L64)


febuiles:
Will take a look at the dependencies to see what's up. With regards to invalid licenses: Let's raise an error if a non SPDX-compliant license is found is the configuration lists (we'll need to bump the major version after this).

cnagadya:
@febuiles this [PR already raises an error](https://github.com/actions/dependency-review-action/blob/79da0131452e55c28c979ebbc97ae8a45fe1efc2/src/config.ts#L33-L47) for non SPDX-compliant licenses in the config lists . Was wondering what should be done for changes with a package whose license does not meet the SPDX spec.  Should `getDeniedLicenseChanges`' [return value](https://github.com/actions/dependency-review-action/blob/57b4a38caa1e0da03150722eb011779ac73003b6/src/licenses.ts#L71) be modified to include non compliant licenses. For now, I've just sneaked in[ console.log](https://github.com/actions/dependency-review-action/blob/57b4a38caa1e0da03150722eb011779ac73003b6/src/licenses.ts#L58) for such cases since `spdx-satisifies` [through spdx-expression-parse.js would throw an error](https://github.com/jslicense/spdx-expression-parse.js/blob/master/scan.js#L122-L126)

cnagadya:
@febuiles, [the error ](https://github.com/jslicense/spdx-expression-parse.js/blob/master/scan.js#L122-L126)is not the most intuitive. It is simply 

> Unexpected `B` at offset 0

febuiles:
@cnagadya can we add a `try` around the calls that can raise the error and wrap it up in something more usable?

cnagadya:
I'm already doing that @febuiles. The question is more about  if we want to surface such packages to the user hence adding a new section to forbidden (currently invalid) and null licenses or just logging them somewhere. Not sure yet if we have a logger setup in the action though

cnagadya:
I think the only case this would happen is for packages with null licenses fetched from the GH endpoint. I remember seeing the spdx_id set to _NOASSERTION_ or something along those lines

febuiles:
> The question is more about if we want to surface such packages to the user hence adding a new section to forbidden (currently invalid) and null licenses or just logging them somewhere.

I think it makes sense to add a new section inside licenses to call out specifically that the license is invalid for SPDX. Thanks for calling this out. 

In terms of logging: We have access to the Actions `core` library [logging utilities](https://github.com/actions/toolkit/blob/main/packages/core/README.md#logging).

> I think the only case this would happen is for packages with null licenses fetched from the GH endpoint. I remember seeing the spdx_id set to NOASSERTION or something along those lines

Null licenses coming from GH API should be listed as "No license found" instead of "Invalid license". I thought `NOASSERTION` was accepted under SPDX, if it's not then we might need to deal with that case manually.

README.md
```diff
@@ -106,37 +106,35 @@ fail-on-scopes:
 
 ### allow-licenses
 
-Only allow the licenses in this list. See "[Licenses](https://github.com/actions/dependency-review-action#licenses)".
+Only allow the licenses that comply with the expressions in this list. See "[Licenses](https://github.com/actions/dependency-review-action#licenses)".
```
febuiles, line 109:
```suggestion
Only allow the licenses in this list. The license name(s) should be SDPX-compliant. See "[Licenses](https://github.com/actions/dependency-review-action#licenses)".
```

Small reword, justification: if my repo contains code/assets under two licenses then my repo's/project's license is `MIT AND GPL-2`. Although we refer to this as an "expression" when parsing, for the end user every one of these expressions is a license by itself. A user should not be writing SPDX expressions by hand since likely no licenses in the SPDX database will match custom made patterns.

README.md
```diff
@@ -106,37 +106,35 @@ fail-on-scopes:
...
 
-**Possible values**: Any `spdx_id` value(s) from
-https://docs.github.com/en/rest/licenses.
+**Possible values**: Any valid [spdx license expression](https://spdx.dev/spdx-specification-21-web-version/#h.jxpfx0ykyb60).
```
febuiles, line 111:
```suggestion
**Possible values**: A list of of [SPDX-compliant license identifiers](https://spdx.org/licenses/).
```

This URL contains a list of the allowed expressions by listing them instead of forcing users to read the entire spec 😅

README.md
```diff
@@ -106,37 +106,35 @@ fail-on-scopes:
...
-**Inline example**: `allow-licenses: BSD-3-Clause, MIT`
+**Inline example**: `allow-licenses: BSD-3-Clause, LGPL-2.1 OR MIT OR BSD-3-Clause`
 
 **YAML example**:
 
 ```yaml
 allow-licenses:
   - BSD-3-Clause
-  - MIT
+  - LGPL-2.1 OR MIT OR BSD-3-Clause
```
febuiles, line 120:
```suggestion
  - LGPL-2.1
  - MIT
  - BSD-3-Clause
```

We should not encourage users to write (or know/care) about SPDX expressions.

febuiles (commented):
First pass of comments, leaving a few suggestions!

src/summary.ts
```diff
@@ -1,18 +1,21 @@
...
 ): void {
   core.summary
     .addHeading('Dependency Review')
-    .addRaw(
-      `We found ${addedPackages.length} vulnerable package(s), ${licenseErrors.length} package(s) with incompatible licenses, and ${unknownLicenses.length} package(s) with unknown licenses.`
-    )
+    .addRaw('We found:')
+    .addList([
+      `${addedPackages.length} vulnerable package(s)`,
+      `${invalidLicenseChanges.unresolved.length} package(s) with unprocessable licenses`,
```
febuiles, line 15:
```suggestion
      `${invalidLicenseChanges.unresolved.length} package(s) with invalid SPDX license definitions`,
```

src/summary.ts
```diff
@@ -93,62 +95,59 @@ export function addLicensesToSummary(
...
-          renderUrl(change.source_repository_url, change.name),
-          change.version
-        ])
-      }
-
-      core.summary.addTable([['Package', 'Version'], ...rows])
-    }
+function formatLicense(license: string | null): string {
+  if (license === null || license === 'NOASSERTION') {
+    return 'Null'
```
febuiles, line 148: 😄

febuiles (approved):

PR #306, opened 2022-10-26
Read configuration from external repositories

### LINKED ISSUE #285: QUESTION: Config file outside current repo
I've tried to update our Dependency Review Action to use a config file, but trying to use a config file in our org-wide .github repo (see: https://github.com/10up/insert-special-characters/pull/164).  However the error I'm seeing in that [action run](https://github.com/10up/insert-special-characters/actions/runs/3236838774) seems to point to the config file needing to be within the report.  Can you clarify if the config file can be in a different repo and if so what the correct format is for that?  Thanks!

cnagadya:
This PR adds support for reading configuration options from files external to the repo where the Action is being run. This enables users to have the config file in another repository, allowing organizations to centralise the configuration in a single file.  

Closes #285.


src/config.ts
```diff
@@ -47,19 +47,28 @@ function validateLicenses(
...
+
+  if (repoConfigFile !== undefined) {
+    const fileContents = readConfigFile(getRepoConfig(repoConfigFile))
+    repoConfig = {...repoConfig, ...fileContents}
   }
+  // the reasoning behind reading the inline config when an external
+  // config file is provided is that we still want to allow users to
+  // pass inline options in the presence of an external config file.
+  // TO DO check order of precedence
+  return {...inlineConfig, ...remoteConfig, ...repoConfig}
```
cnagadya, line 71: What is the correct order of precedence here?

src/utils.ts
```diff
@@ -38,3 +40,16 @@ export function isSPDXValid(license: string): boolean {
     return false
   }
 }
+
+export function octokitClient(token = 'repo-token', required = true): Octokit {
```
cnagadya, line 44: keeping this here for now but maybe worth it moving it a separate file. Any reservations using the same client instead of [this](https://github.com/actions/dependency-review-action/blob/main/src/dependency-graph.ts#L6-L9) in `dependency-graph.ts` [since retrying once is enabled by default](https://github.com/octokit/octokit.js/blob/main/README.md#constructor-options)?
image

__tests__/config.test.ts
```diff
@@ -93,134 +96,134 @@ test('it raises an error when no refs are provided and the event is not a pull r
...
-  let options = readConfig()
+  let options = await readConfig()
   expect(options.allow_licenses).toEqual(undefined)
   expect(options.deny_licenses).toEqual(undefined)
 
   setInput('config-file', './__tests__/fixtures/license-config-sample.yml')
-  options = readConfig()
+  options = await readConfig()
   expect(options.fail_on_severity).toEqual('low')
 })
```
cnagadya, line 141: Still WIP. Currently skipped to stop ci from breaking but need updating due to introduction of loading an external config file

src/config.ts
febuiles: We should maintain the same precedence we have today (`Object.assign({}, inlineConfig, config)`, while **allowing the users to overwrite values**.

This means that if an external/remote config file specifies a value "A", and a user overwrites that in their local repo with value "B", we'll choose "B" as the definitive answer.

__tests__/config.test.ts
```diff
@@ -171,8 +172,8 @@ test('it raises an error when given invalid scope', async () => {
   setInput('fail-on-scopes', 'runtime, zombies')
   await expect(readConfig()).rejects.toThrow(/received 'zombies'/)
 })
-
-test('it defaults to an empty GHSA allowlist', async () => {
+//TO DO: Undefined !== []. Clarify test intent
```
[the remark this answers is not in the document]
febuiles: I'm not sure if I wanted to confirm anything besides the fact that by default no GHSAs are filtered? :|

febuiles:
@courtneycl would you mind helping us review the copy for `README.md`? We could use a fresh set of eyes :)

febuiles (approved):

README.md
```diff
@@ -71,13 +71,20 @@ or by inlining these options in your workflow file.
...
-configuration file.
+If the configuration file is located in an external private repository,
+use the [external-repo-token](#external-repo-token) parameter of the
+Action to specify a token that has read access to the repository.
 
-**Example**: `config-file: ./.github/dependency-review-config.yml`.
+**Possible values**: A string representing an absolute path to a file,
+or a file located in another repository:
+
+**Example**: `config-file: ./.github/dependency-review-config.yml # local file`.
```
courtneycl, line 85: my dumb question, I thought this is a relative path? 🤔

courtneycl (commented):
Made just a couple suggestions. I do think we should revamp this and go with a table or something to make it easier to read; I'll take a whack at that in another PR after this one.

README.md
febuiles: Not dumb at all, the example is indeed a relative path! I wrote the docs before we implemented any functionality, will change this!

febuiles: Pushed a fix in 5ea8fbf, don't love it but it's correct. Feel free to change it in the next README PR 😄

febuiles:
@courtneycl I made a few tweaks to the README, feel free to open a new PR to update the documentation.

PR #311, opened 2022-10-28
Add `license-check` and `vulnerability-check` inputs

### LINKED ISSUE #308: Support disabling vulnerability or license check
Suggestion to add support to disable either the vulnerability or license check. From the user perspective, something along the lines of:

```yaml
- name: Dependency Review
  uses: actions/dependency-review-action@v2
  with:
    dependency-check: false   # `true` to enable, `false` to disable
    license-check: true       # `true` to enable, `false` to disable
```

I believe this is useful when you already use a different tool to do one of these checks. For example, say I use [`npm audit`](https://docs.npmjs.com/cli/v8/commands/npm-audit) for vulnerability scanning and I need a tool for license checking. In this scenario, using this Action is possible but I don't really need the dependency check. Moreover, it would likely lead to duplicate alerts, which can get noisy. Hence, I would want to disable the vulnerability scanning. One can imagine the same could be true for the license check.

#184 looks related as it would allow you to make only one of the two checks required. I still believe completely disabling one of the two checks would be a preferable solution in order to reduce noise.

ericcornelissen:
Closes #308

This add support for and document two new inputs:

- `license-check`: Can be used to disable all license-related checks performed by this action.
- `vulnerability-check`: Can be used to disable all vulnerability-related checks performed by this action.

A check is included that prevents both checks to be disabled since this would render this action useless.


__tests__/config.test.ts
```diff
@@ -181,6 +183,46 @@ test('it successfully parses GHSA allowlist', async () => {
...
+  let options = readConfig()
+  expect(options.vulnerability_check).toEqual(false)
+
+  clearInputs()
+  setInput('vulnerability-check', 'true')
+  options = readConfig()
+  expect(options.vulnerability_check).toEqual(true)
+})
+
+test('it is not impossible to disable both checks', async () => {
```
febuiles, line 218:
```suggestion
test('it is not possible to disable both checks', async () => {
```

febuiles (approved):

febuiles:
@ericcornelissen This looks great, thanks a ton! The changes will go into the next release of the Action (sometime this week).

PR #330, opened 2022-11-15
Improve error messages for external config files

febuiles:
We're currently showing the wrong error message (can't fetch config file) whenever the external config file cannot be parsed. This issue was surfaced in this [PR comment](https://github.com/actions/dependency-review-action/issues/285#issuecomment-1314822987).

This PR corrects the behavior of comma-separated lists in both inline and external config files, it also updates the error messages to move closer to the user expectations.


src/config.ts
```diff
@@ -123,7 +129,14 @@ function parseConfigFile(configData: string): ConfigurationOptionsPartial {
     const data = YAML.parse(configData)
     for (const key of Object.keys(data)) {
       if (key === 'allow-licenses' || key === 'deny-licenses') {
-        validateLicenses(key, data[key])
+        const licenses = data[key]
+        // handle the case where the user has a singe or invalid license
+        // in the config file
+        if (typeof licenses === 'string') {
+          validateLicenses(key, [licenses])
```
cnagadya, line 136: Is it intended to treat the whole provided value as one license? This might be confusing in the case of the reported issue. Or maybe we should just throw an error for non list values provided 🤔

febuiles: The original comment had this structure for an external config file:

```yaml
allow-licenses: FOO, BAR, BAZ, BLA
```

In YAML that is equal to `allow-licenses: "FOO, BAR, BAZ, BLA"` which is indeed one license.

I think the confusion stems from the fact that in the inline configuration you can provide a list by using commas (GitHub Actions don't support array-like inputs), while the external config file expect YAML lists.

Now that you bring this up I'm noticing that I'm likely blowing up our inline license parsing with this change and I need to make some changes, right? 😅

hmaurer: > I think the confusion stems from the fact that in the inline configuration you can provide a list by using commas (GitHub Actions don't support array-like inputs), while the external config file expect YAML lists.

@febuiles I think that's definitely a problem; we should not have a discrepancy between the inline configuration and the external configuration. Is there anything stopping us from parsing the external configuration in the same way as we parse the inline configuration, while also allowing YAML arrays?

febuiles: @hmaurer I just did a quick read of the [SPDX license names guidelines](https://github.com/spdx/license-list-XML/blob/main/DOCS/license-fields.md) and they specifically ask not to use commas in the full name/ID. I think supporting the same format with commas for both config options is what makes sense here.

src/config.ts
```diff
@@ -80,12 +80,11 @@ function validateLicenses(
   if (licenses === undefined) {
     return
   }
+
   const invalid_licenses = licenses.filter(license => !isSPDXValid(license))
 
   if (invalid_licenses.length > 0) {
-    throw new Error(
-      `Invalid license(s) in ${key}: ${invalid_licenses.join(', ')}`
```
febuiles, line 87: I removed the formatting since it was making it hard to debug failures. This resulted in a couple of changes in the tests where we perform string comparisons.

febuiles:
@cnagadya @hmaurer I've updated the PR to add support for inline config lists, would appreciate another review.

src/config.ts
febuiles: This went from being a "Fix an error message" PR to "Fix lists and also an error messages". I've updated the PR description and cleaned up the code a bit to account for the cases where we have comma-separated lists in external configs.

cnagadya (approved):
Awesome, great job also adding the other listy fields 🥳

PR #331, opened 2022-11-15
Set octokit baseurl for GHES

cnagadya:
As stated in the title, this set the [required baseUrl](https://github.com/octokit/octokit.js/blob/9c8fa89d5b0bc4ddbd6dec638db00a2f6c94c298/README.md?plain=1#L196) for GHES in the octokit client


src/utils.ts
```diff
@@ -41,6 +41,13 @@ export function isSPDXValid(license: string): boolean {
   }
 }
 
+function isEnterprise(): boolean {
+  const serverUrl = new URL(
+    process.env['GITHUB_SERVER_URL'] ?? 'https://github.com'
```
cnagadya, line 46: GITHUB_SERVER_URL is [always set](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) in the Action Environment Variables but process.env returns string | undefined

febuiles (approved):

PR #370, opened 2023-01-05
Fix Dependency Review API response error handling

felickz:
Bug: Failures in the Dependency Review API response are not properly handled due to an octokit version mismatch 
- To reproduce: Invoke Dependency Review on a repo without Dependency graph or a private repo without GitHub Advanced Security to create a 403 response from the API.  Error handling falls back into the default catch block as exception instance type checking fails based on a breaking format change
image

Fixes:
- downgrade octokit/request-error to ^2.1.0
   - supported by [actions/core ^1.10.0](https://github.com/actions/dependency-review-action/blob/90014ebf46221852a6aa4dcbce7f08daf4b9e4c5/package.json#L28) ([see similar issue](https://github.com/octokit/request-error.js/issues/246))
- Add integration test to invoke API (to avoid mocking a response that is dependent on the version of `actions/core` used) and validate exception response
- Enhanced `Forbidden` error message back to user to hint that a culprit might also be lack of GitHub Advanced Security for a private repo

## Graph Disabled / Private Repo w/o GHAS
Before:
```
::error::Forbidden
```
image

After:
```
::error::Dependency review is not supported on this repository. Please ensure that Dependency graph is enabled along with GitHub Advanced Security on private repositories, see https://github.com/OWNER/REPO/settings/security_analysis
```

image


felickz:
Integration Test fails with 3.0.2
image

Succeeds with 2.1.0
image

elireisman (commented):
This fix and explanation on the PR description LGTM, thanks! ❤️ 

@febuiles @hmaurer any thoughts on the prospect of us filing a bug and rolling forward with the `RequestError` format change and upgrade of Octokit `request-error` vs downgrade presented here? Or other blocking concerns?

Otherwise, this seems ready to merge 👍

febuiles:
Thank you for this @felickz, with the new integration test I'm sure we'll be prevented of breaking this behavior by auto-updating the errors library 😅. 
This changes will appear in 3.0.3.

PR #439, opened 2023-03-22
Show snapshot warnings in the summary

juxtin:
The compare endpoint will now sometimes return a header called `x-github-dependency-graph-snapshot-warnings`, which contains additional information about snapshot comparisons that users should know about. Right now, this header is only returned if the repository is flagged in for snapshot comparisons.

This PR adds code to look for the header, decode it if it is present, and display it in the summary. The text is sort of beta at this point, and should be updated when we have a real documentation link.

### Random bugfix

I also added a bonus bugfix! Sometimes snapshot detectors don't add a manifest path for some reason and just use `""` instead. Maybe we shouldn't allow this, but it happens now and it does funky things in the HTML summary:

Screenshot 2023-03-24 at 12 07 35 PM

As you can see in the screenshot, the fold should be closed, but you can still see all the dependencies spilling out of it. On this branch, the fold works:

Screenshot 2023-03-24 at 12 22 33 PM

Closes https://github.com/github/dependency-graph/issues/1898


juxtin:
Here's a screenshot of this in action.

Screenshot 2023-03-22 at 2 39 51 PM

__tests__/summary.test.ts
```diff
@@ -27,6 +27,45 @@ const defaultConfig: ConfigurationOptions = {
...
+    version: '0.1.17',
+    package_url: 'pkg:hex/castore@0.1.17',
+    license: null,
+    source_repository_url: null,
+    scope: 'runtime',
+    vulnerabilities: []
+  },
+  {
+    change_type: 'added',
+    manifest: '',
```
lseppala, line 45: ❤️ thanks for capturing this invariant in a test!

src/utils.ts
```diff
@@ -8,7 +8,9 @@ export function groupDependenciesByManifest(
 ): Map<string, Changes> {
   const dependencies: Map<string, Changes> = new Map()
   for (const change of changes) {
-    const manifestName = change.manifest
+    // If the manifest is null or empty, use a space as the key to avoid
+    // breaking the HTML rendering later
+    const manifestName = change.manifest || ' '
```
lseppala, line 13: Rather than using invisible characters that could be confusing, maybe let's consider something like `'<no manifest specified>'` to make it clearer to the user?

lseppala (approved):
Really nice work. I have one product-changing request to help make it clearer when no manifest is specified, but this can be a fast-follow PR.

PR #451, opened 2023-04-06
Fix default values for fail-on-severity

### LINKED ISSUE #422: remote config file setting not getting set/used
Setting up dependancy-review to use an external repo for its config.  
I can't get the config file settings to actually be used/applied
I have confirmed that the repo is accessable and the config file is being loaded by changing the name of the config file to something that doesn't exist and got an error: "Error: Unable to fetch or parse config file: Error fetching remote config file"
Putting a bad, yaml syntax also generates an error:
`Unable to fetch or parse config file: Implicit keys need to be on a single line at line 3, column 1: # Possible values: "critical", "high", "moderate", "low" this-is-notused-bad ^`

But no matter what setting I put in the config.yml dependancy-review always reports using its default settings, "fail-on-severity: low", "fail-on-scopes: runtime" not the settings in the config file.

Config that we are testing with, remote repo with the following config:
```
fail-on-severity: 'critical'
fail-on-scopes:
  - development
  - runtime
```

Repo that will be using the above config:
```
name: 'Dependency Review'
on: [pull_request]

permissions:
  contents: read

jobs:
  dependency-review:
    runs-on: ubuntu-latest
    steps:
      - name: 'Checkout Repository'
        uses: actions/checkout@v3
      - name: 'Dependency Review'
        uses: actions/dependency-review-action@v3
        with:
          config-file: 'org/sec-github-actions-config/dependancy-review/config.yml@master'
          external-repo-token: ${{ secrets.SEC_DEPENDANCY_REVIEW }}
```

But when the action runs it always reports the following in the action:
```
Run actions/dependency-review-action@v3
  with:
    config-file: org/sec-github-actions-config/dependancy-review/config.yml@master
    external-repo-token: ***
    repo-token: ***
    fail-on-severity: low
    fail-on-scopes: runtime
```
So my custom setting are not being used/applied but I get no error loading/parsing the file.  Running the action in debug mode didn't give any extra info/errors.

I'm testing with a Pipfile that only has a 'high' issue in it related to the lxml package.

```
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[[source]]

[dev-packages]

[packages]
python-json-logger = "==2.0.1"
lxml = "==4.6.1"
boto3 = "*"

[requires]
python_version = "3.8"

```

Any ideas on what am I missing/doing wrong?

febuiles:
Closes #422.

Today the Action is always returning the results as if the user had set `fail-on-severity` to `low` when using an external configuration file. This issue arises from having default values in [action.yml](https://github.com/actions/dependency-review-action/blob/main/action.yml#L12). When the Action runs it will see that no value was set for a field, and since it has a `default` it can use, it inserts it, making our code think that this was a user-specified inline config option.

Most of this PR adds tests, documentation and moves things around. The actual fix is in 9885d0c74cca15bb0f8128ee7a4018cbac4d6c32. 

PS: While working on this I realized that at some point our `README` was updated to include options with dashes in the names instead of underscores (`fail-on-severity` vs `fail_on_severity`). I would love if we supported both, but at the moment only the latter format is supported in external files. I'm highlighting this in this PR although the changes are unrelated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,8 @@ Start by specifying that you will be using an external configuration file:
     config-file: './.github/dependency-review-config.yml'
 ```
 
-And then create the file in the path you just specified:
+And then create the file in the path you just specified. Please note
+that the **option names in external files use underscores instead of dashes**:
 
 ```yaml
 fail_on_severity: 'critical'
--- a/README.md
+++ b/README.md
@@ -131,8 +131,8 @@ Start by specifying that you will be using an external configuration file:
 And then create the file in the path you just specified:
 
 ```yaml
-fail-on-severity: 'critical'
-allow-licenses:
+fail_on_severity: 'critical'
+allow_licenses:
   - 'GPL-3.0'
   - 'BSD-3-Clause'
   - 'MIT'
--- a/__tests__/external-config.test.ts
+++ b/__tests__/external-config.test.ts
@@ -1,35 +1,7 @@
 import {expect, test, beforeEach} from '@jest/globals'
 import {readConfig} from '../src/config'
 import * as Utils from '../src/utils'
-
-// GitHub Action inputs come in the form of environment variables
-// with an INPUT prefix (e.g. INPUT_FAIL-ON-SEVERITY)
-function setInput(input: string, value: string): void {
-  process.env[`INPUT_${input.toUpperCase()}`] = value
-}
-
-// We want a clean ENV before each test. We use `delete`
-// since we want `undefined` values and not empty strings.
-function clearInputs(): void {
-  const allowedOptions = [
-    'FAIL-ON-SEVERITY',
-    'FAIL-ON-SCOPES',
-    'ALLOW-LICENSES',
-    'DENY-LICENSES',
-    'ALLOW-GHSAS',
-    'LICENSE-CHECK',
-    'VULNERABILITY-CHECK',
-    'CONFIG-FILE',
-    'BASE-REF',
-    'HEAD-REF',
-    'COMMENT-SUMMARY-IN-PR'
-  ]
-
-  // eslint-disable-next-line github/array-foreach
-  allowedOptions.forEach(option => {
-    delete process.env[`INPUT_${option.toUpperCase()}`]
-  })
-}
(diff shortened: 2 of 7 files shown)


__tests__/external-config.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  )
+  const config = await readConfig()
+
+  expect(config.allow_licenses).toEqual(['MIT', 'GPL-2.0-only'])
+})
+
+test('it reads a config file hosted in another repo', async () => {
+  setInput(
+    'config-file',
+    'future-funk/anyone-cualkiera/external-config.yml@main'
```
elireisman, line 103: 🤔 I wonder if we should put this file into a static repo under the `actions` org or something?

README.md
```diff
@@ -128,11 +128,12 @@ Start by specifying that you will be using an external configuration file:
     config-file: './.github/dependency-review-config.yml'
 ```
 
-And then create the file in the path you just specified:
+And then create the file in the path you just specified. Please note
+that the **option names in external files use underscores instead of dashes**:
```
elireisman, line 132: 🤔 I assume there's some known friction involved 👍  but I wonder could we relax the option name parsing to accept both formats? Feels like this will be tough for end users to remember

__tests__/external-config.test.ts
febuiles: We are mocking the response to avoid HTTP calls: https://github.com/actions/dependency-review-action/pull/451/files/654eb5ca1c29dea80ae435b7e64ac77ec50eb2da#diff-82828d4b4ae35e47f69929a800f360461e236bea4371942e452bb901ea937c2dR9-R26

README.md
febuiles: Agreed, it's on the TODO. This PR only fixes the bug with defaults value, but going forward I think supporting both options by replacing `-` with `_` when found is the way.

elireisman (approved):
Thanks for answers to the callouts 🙇  sounds like the follow on work needed to complete this one is already on deck. LGTM! 🚀

PR #452, opened 2023-04-06
Show all non-empty snapshot warnings

juxtin:
### Current behavior:

The snapshot warning `No snapshots found for head commit [sha]…` is never displayed, because this warning would appear for users who might not be interested in this feature.

### Behavior with this PR:

All non-empty snapshot warnings are displayed in the summary. Note that only repositories that have canonical snapshots will receive warnings of any kind.


juxtin:
We will be merging this, per conversation with @jhutchings1

mrysav (approved):

elireisman (approved):

Issue #493, opened 2023-05-31
A purl string argument is required.


tamacti-jun:
Hello @febuiles , @theztefan , @cnagadya, @sarahkemi 

![image](https://github.com/actions/dependency-review-action/assets/134579857/72d018a8-ace9-4dc5-82ef-96028fda5691)

hi, I'm getting this error, what should I do to solve the problem?

my file content is as follows.

![image](https://github.com/actions/dependency-review-action/assets/134579857/1a264310-7093-4f49-8c31-856fe986375d)

tamacti-jun:
I don't understand what went wrong. my github repo is a public repo

![image](https://github.com/actions/dependency-review-action/assets/134579857/5f8fa6d2-7082-48e2-97cd-bf0e22962be4)
t

febuiles:
@tamacti-jun Thanks for reporting the bug. We're working on a fix, in the meantime can I suggest you use `dependency-review-action@3.0.4` in your config file?

febuiles:
@tamacti-jun This should be fixed in `v3` and `v3.0.6` of the Action, thanks for the report.

tamacti-jun:
Thank you for your interest. have a nice work day

PR #494, opened 2023-05-31
Empty PURLs should not block the action from running

febuiles:
Fixes https://github.com/actions/dependency-review-action/issues/493.


theztefan (approved):
This should fix the #493. 

It will still fail when the `package_url` is not a valid purl string, but given we get this information from Dependancy Graph (API) this should not happen/be possible.

Issue #538, opened 2023-08-02
Comment on PR Only if Failure


henry-prince-addepar:
It would be incredibly helpful and a better developer experience if you could add an option for `comment-summary-in-pr` that only comments `on-failure` and otherwise stays out of the way for passing checks. Thanks!

febuiles:
@henry-prince-addepar that's a good idea! I don't have the bandwidth to implement this atm, but I'm leaving this open in case someone else is interested! If you want to submit a pull request I'd be happy to help, just ping me!

PR #540, opened 2023-08-05
Add `on-failure` option to `comment-summary-in-pr` setting

sgmurphy:
This PR resolves https://github.com/actions/dependency-review-action/issues/538 by changing `comment-summary-in-pr` from a boolean to an enum of `always`, `never`, `on-failure`. When `on-failure` is used the summary will only be posted in a comment if the workflow status is 'failure'.


febuiles:
@sgmurphy This idea looks good, thank you 👍. It will existing behavior (`true`/`false` values) in order to keep supporting users. I'm not sure how you'd like to proceed, an idea:

1. We check the input value and if it's a boolean we map `true`/`false` to the new string values.
2. Update the docs with a mention to this change.
3. We do a new release bumping the minor version number to 3.1, minimizing impact to existing users with pinned versions.

I can help with (2) and (3), but would like your input on (1), how do you think we should handle this transition for existing users without breaking their existing setups?

sgmurphy:
@febuiles Thanks for the feedback. I updated the PR to map `true` => `'always'` and `false` => `'never'` so this change shouldn't impact existing users. I also added some unit tests to confirm the mapping works as expected.

febuiles:
@sgmurphy the new options are available starting with [v3.0.8](https://github.com/actions/dependency-review-action/releases/tag/v3.0.8), thank you very much!

uni-manojkumar-n:
Hey @sgmurphy 
like my doubt is I didn't get it 
we have same requirement to implement this in our CICD pipeline so can you please explain it briefly

PR #556, opened 2023-08-31
Dr Snaps launch PR

juxtin:
This PR includes the following prerequisites for the Dr Snaps launch:
* the `retry-on-snapshot-warnings` and `retry-on-snapshot-warnings-timeout` config options, courtesy of @hmaurer 
* the removal of filtering snapshot warnings, brought in from https://github.com/actions/dependency-review-action/pull/452.
* setting `retry-on-snapshot-warnings` to `false` by default.
* adding a link to the (forthcoming) docs when displaying snapshot warnings.


jhutchings1:
Dr Snaps sounds like a cool cartoon doctor. 

Also, thanks for pulling all this together @juxtin! ✨

src/dependency-graph.ts
```diff
@@ -28,7 +28,7 @@ export async function compare({
   const changes = await octo.paginate(
     {
       method: 'GET',
-      url: '/repos/{owner}/{repo}/dependency-graph/compare/{basehead}',
+      url: '/repos/{owner}/{repo}/dependency-graph/compare/{basehead}?includes_dependency_snapshots=true',
```
juxtin, line 31: This param is unnecessary now, it's controlled by the feature flag.

```suggestion
      url: '/repos/{owner}/{repo}/dependency-graph/compare/{basehead}',
```

src/main.ts
```diff
@@ -18,18 +18,60 @@ import {groupDependenciesByManifest} from './utils'
...
+}
+
+async function getComparison(
+  baseRef: string,
+  headRef: string,
+  retryOpts?: {
+    retryUntil: number
+    retryDelay: number
+  }
+): ReturnType<typeof dependencyGraph.compare> {
```
febuiles, line 32: (Q) why do we do this instead of specifying an explicit type?

febuiles (approved):
LGTM. Once we've clarified what the default value should be in `action.yml` I think we're good to go.

action.yml
```diff
@@ -53,6 +53,14 @@ inputs:
   deny-groups:
     description: A comma-separated list of package URLs for group(s)/namespace(s) to deny (e.g. "pkg:npm/express, pkg:pip/pycrypto")
     required: false
+  retry-on-snapshot-warnings:
+    description: Whether to retry on snapshot warnings
+    required: false
+    default: true
```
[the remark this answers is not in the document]
juxtin: oh thanks, I missed this. I changed the default in the typescript code but not in this file.

src/main.ts
juxtin: My guess is that it's because this is just a wrapper function, so passing through the type as well as the value could simplify things when refactoring.

febuiles (approved):

PR #578, opened 2023-09-29
Updated example permissions in README.md

ctcampbell:
The permissions detailed in the readme will result in a 403 from the dependency snapshot submission API if this action is used in conjunction with e.g. `advanced-security/maven-dependency-submission-action`. It should be `contents: write` to resolve this, so I have added a couple of comments to the readme as such.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,9 @@ on: [pull_request]
 
 permissions:
   contents: read
+  # If using a dependency submission action such as advanced-security/maven-dependency-submission-action
+  # this permission must be set to:
+  # contents: write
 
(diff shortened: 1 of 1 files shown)


juxtin:
Hi @ctcampbell, thanks for the contribution! You are correct, `contents: write` is required for dependency submission actions. That said, I'm hesitant to merge this. _This_ action doesn't require those permissions, so it seems like it's probably not the best place to document that requirement.

febuiles:
Closing this given that the behavior is already documented in the linked Action README.

PR #639, opened 2023-12-14
Update action to node20

takost:
**Description:**
Node 16 has reached end-of-life on 11 Sep 202.
This PR updates the default runtime to node20, rather then node16.

This is supported on all Actions Runners [v2.308.0](https://github.com/actions/runner/releases/tag/v2.308.0) or later.

### DIFF
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -23,10 +23,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set Node.js 18.x
+      - name: Set Node.js 20.x
         uses: actions/setup-node@v4
         with:
-          node-version: 18.x
+          node-version: 20.x
           cache: npm
 
       - name: Install dependencies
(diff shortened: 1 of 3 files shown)


takost:
@febuiles Could you please help me to test it properly? I am not sure how to reproduce that problem with proxy, so not sure about testing. Also I think we can release new major version of this action, for other first party actions we released new major versions after node change

febuiles:
@takost Apologies for the delay getting back to you, I was out visiting family.

In order to test the changes, we need to setup a self-hosted runner (can be done on your local machine) and the add the  HTTPS proxy in front of it. (`nc` and `stunnel` should do the job, or something like `nginx` if you already have that installed).

> I think we can release new major version of this action, for other first party actions we released new major versions after node change

Let's do this. Do you mind bumping it in this PR? We can get this merged so you have confirmation that `main` is running Node 20. I can perform the testing somewhere in the next two weeks and update the docs to point to `v4`

takost:
@febuiles I tested the PR with proxy behind runner, and my workflow passed.

febuiles (approved):
@takost thank you very much!

PR #671, opened 2024-01-25
Create stale.yaml

jonjanego:
Creating a stalebot workflow that will check every day for stale PRs older than 180 days, then close any that were labeled 'stale' after another 15 days.

It exempts any PRs assigned to the dependency graph dev team.

This will also need a PAT created, and stored as a repo secret with the name ```STALE_BOT```

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,5 +1,8 @@
 name: Close stale PRs
 
+permissions:
+  pull-requests: write
+
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


elireisman (approved):

jonjanego:
> This will also need a PAT created, and stored as a repo secret with the name `STALE_BOT`

updated the workflow definition to use the default github token, with explicit scopes to write PRs

mrysav (approved):
Looks good to me!

PR #678, opened 2024-01-31
Use manual codeql config

juxtin:
We now have CodeQL running on this repository, but it's reporting failures in the generated JavaScript. AFAICT, these alerts are not based on any TypeScript code that we can actually change, so I think the right move is to exclude the generated JavaScript files from Code Scanning and rely on the TypeScript coverage to generate alerts that are actually actionable.

I based this config on the auto-generated one, but I removed a few things like Swift and Ruby matrix items and MacOS builds.

Ultimately the whole point of this is really the line `paths-ignore: dist/index.js,dist/sourcemap-register.js`.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,45 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  schedule:
+    - cron: '21 0 * * 4'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: 'ubuntu-latest'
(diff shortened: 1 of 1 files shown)


juxtin:
Looks like my config wasn't quite right:

```
Warning: Unexpected input(s) 'paths-ignore', valid inputs are ['tools', 'languages', 'token', 'registries', 'matrix', 'config-file', 'db-location', 'config', 'queries', 'packs', 'external-repository-token', 'setup-python-dependencies', 'source-root', 'ram', 'threads', 'debug', 'debug-artifact-name', 'debug-database-name', 'trap-caching']
```

elireisman (approved):
Nice fix! 🚀

PR #695, opened 2024-02-19
Expose dependency comment content

jsoref:
Closes #250

PR #696, opened 2024-02-19
Expose dependency comment content

febuiles:
Tweaks to #695 contributed by @jsoref.


README.md
```diff
@@ -1,21 +1,21 @@
...
 This action scans your pull requests for dependency changes, and will
-raise an error if any vulnerabilities or invalid licenses are being introduced. The action is supported by an [API endpoint](https://docs.github.com/en/rest/reference/dependency-graph#dependency-review) that diffs the dependencies between any two revisions on your default branch.
+raise an error if any vulnerabilities or invalid licenses are being introduced. The action is supported by an [API endpoint](https://docs.github.com/rest/dependency-graph/dependency-review) that diffs the dependencies between any two revisions on your default branch.
 
 The action is available for all public repositories, as well as private repositories that have GitHub Advanced Security licensed.
 
 You can see the results on the job logs:
 
-<img width="854" alt="Screen Shot 2022-03-31 at 1 10 51 PM" src="https://user-images.githubusercontent.com/2161/161042286-b22d7dd3-13cb-458d-8744-ce70ed9bf562.png">
+<img width="850" alt="GitHub workflow run log showing Dependency Review job output" src="https://user-images.githubusercontent.com/2161/161042286-b22d7dd3-13cb-458d-8744-ce70ed9bf562.png">
```
febuiles, line 10: Added a new image, thanks for the feedback!

README.md
```diff
@@ -1,21 +1,21 @@
...
 
 or on the job summary:
 
-<img src="https://user-images.githubusercontent.com/7847935/182871416-50332bbb-b279-4621-a136-ca72a4314301.png">
+<img width="850" alt="GitHub job summary showing Dependency Review output" src="https://github.com/actions/dependency-review-action/assets/2161/42fbed1d-64a7-42bf-9b05-c416bc67493f">
```
febuiles, line 14: Added a new image, thanks for the feedback!

README.md
```diff
@@ -1,21 +1,21 @@
 # dependency-review-action
 
 This action scans your pull requests for dependency changes, and will
-raise an error if any vulnerabilities or invalid licenses are being introduced. The action is supported by an [API endpoint](https://docs.github.com/en/rest/reference/dependency-graph#dependency-review) that diffs the dependencies between any two revisions on your default branch.
+raise an error if any vulnerabilities or invalid licenses are being introduced. The action is supported by an [API endpoint](https://docs.github.com/rest/dependency-graph/dependency-review) that diffs the dependencies between any two revisions on your default branch.
```
febuiles, line 4: Changed this to point directly to the DR API.

docs/examples.md
```diff
@@ -164,6 +164,39 @@ jobs:
...
+
+Using the `comment-content` output you can get the results of the action in a workflow step.
+
+```yaml
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+  pull-requests: write
```
febuiles, line 177: This is necessary to get the nice comment output on the PRs.

docs/examples.md
```diff
@@ -164,6 +164,39 @@ jobs:
...
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v4
+      - name: 'Dependency Review'
+        id: review
+        uses: actions/dependency-review-action@v4
+        with:
+          fail-on-severity: critical
+          deny-licenses: LGPL-2.0, BSD-2-Clause
+      - name: 'Report'
+        if: always()  # make sure this step runs even if the previous failed
```
febuiles, line 192: Had to add this to make sure this step was run regardless of the exit status of the previous one. When DR Action was run on a pull request and it failed it would stop execution (sample run here).

I have validated this workflow here: https://github.com/future-funk/bookish-eureka/pull/4. See the details for the `echo` output.

docs/examples.md
jsoref: Oh, my goal was to *not* have a comment. Can we instead include the setting to suppress the normal comment logic?

Maybe the workflow author only wants to use a job summary

docs/examples.md
febuiles: @jsoref Gonna run with this example for the release, if you know of a better way to get around the Action failures please let me know and I'll be happy to update the docs.

docs/examples.md
febuiles: > Can we instead include the setting to suppress the normal comment logic?

Sure, we can remove this permission if that's the case, happy to approve a PR.

docs/examples.md
jsoref: I think we want: [`jobs.<job_id>.steps[*].continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)

febuiles: From the description: `Set to true to allow a job to pass when this step fails`. We'd be marking the run as a success even if it fails, which is not a behavior we want. I'm not too familiar with workflows, so please let me know if I'm misreading this.

jsoref: If the goal is for the job to fail and to generally yield output, then you want: `${{ !cancelled() }}` see the note in https://docs.github.com/en/actions/learn-github-actions/expressions#always

febuiles: Thanks for the pointer. `!cancelled()` skipped the `Report` step: https://github.com/future-funk/bookish-eureka/actions/runs/7963028628/job/21737777423. 

There is a `if: ${{ failure() && steps.review.conclusion == 'failure' }}` condition in there that looks much more appropriate for this case. I was able to validate it's working in https://github.com/future-funk/bookish-eureka/actions/runs/7963058574/job/21737875771?pr=6. I'll update the example!

febuiles: Fixed in https://github.com/actions/dependency-review-action/commit/587ff57.

Issue #250, opened 2022-09-22
Supporting action output


davidmally-at:
👋 Hi! My team has been using this action, and we were wondering if it's possible to get the summary output from the action to use in subsequent workflow steps. A fake example:
![image](https://user-images.githubusercontent.com/61764280/191811701-7209e891-b0fd-4827-8e5b-616977c03e08.png)

We'd like to have this summary output available as output from the action itself, so that we can do things like post it as a comment on the PR that corresponds to the workflow run, notify people in Slack, etc. Is that supported today and I'm just missing it? If not, I would be happy to put together a pull request to support this.

febuiles:
Hi @davidmally-at! I don't think we have anything in place to support the use case you're proposing, a PR would be extremely welcome! 🙇

YSCG1130:
J

uli-f:
I'd also support this feature request.

The summary used for a PR comment is compiled. All that is left to do is output that from the workflow so that it could be picked up by later steps.

jsoref:
If that's all that's desired, then it'd be a pretty easy thing to do right here:
https://github.com/actions/dependency-review-action/blob/80f10bf419f34980065523f5efca7ebed17576aa/src/comment-pr.ts#L16-L23

febuiles:
This functionality is now available in the latest release of the Action (4.1.2) thanks to @jsoref. You can find an [example here](https://github.com/actions/dependency-review-action/blob/main/docs/examples.md#getting-the-results-of-the-action-in-a-later-step) of how to use the output. Closing this issue.

PR #698, opened 2024-02-20
Set comment as output

jsoref:
Move the logic from the caller as a negation instead of trying to negate it by hand.

Second attempt at #250


jsoref:
@febuiles here's a second attempt at the logic. Note that I don't actively use this action. It seemed like a reasonable thing for me to do as a spare time task on my weekend/vacation.

I'm looking carefully and the:
- [x] Allow edits and access to secrets by maintainer

option is set.

I should be generally available if you have questions/need changes.

febuiles:
Thanks! After going through this with fresh morning eyes the conditional makes sense now.

> Note that I don't actively use this action. It seemed like a reasonable thing for me to do as a spare time task on my weekend/vacation.

In this case this contribution is valued twice as much, thanks a ton for your time and help again 🙇

PR #707, opened 2024-03-02
Add outputs for the changes data

laughedelic:
Hi! I'm using this action in conjunction with others and it would be very useful for the subsequent steps to be able to access the data that this action presents in the summary, but in a more structured format. So I added 4 outputs: one with all changes data and 3 optional ones with the data on vulnerable dependencies, invalid licenses and denied dependencies. Here's an example usage:

```yaml
      - name: Dependency review
        uses: actions/dependency-review-action@v4
        id: dependency-review
        with:
          # ...

      - name: Render dependencies lineage
        uses: laughedelic/gha-dependency-lineage@v0
        if: failure()
        with:
          vulnerable-dependencies: ${{ steps.dependency-review.outputs.vulnerable-changes }}
```

The second step is a custom action that analyses dependencies lineage and renders a graph diagram in the job summary. ~~Without these added outputs, an action like that doesn't have any other way to know which dependencies are relevant for the report.~~ (I guess, it's possible to call the [API](https://docs.github.com/en/rest/dependency-graph/dependency-review?apiVersion=2022-11-28#get-a-diff-of-the-dependencies-between-commits) directly, but it's wasteful when the information is already available from this action and it would only provide the unfiltered information).

I would appreciate any feedback and suggestions.


laughedelic:
One thing to consider here is that [action outputs have a size limit](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputs-for-docker-container-and-javascript-actions):

> Outputs are Unicode strings, and can be a maximum of **1 MB**. The total of all outputs in a workflow run can be a maximum of 50 MB.

I don't know how big the dependencies JSON can get. I think 1MB of JSON is a lot 🤔

jonjanego:
Thank you @laughedelic !  I like the idea, but agree with @febuiles that we'd like to see some more examples and documentation.

I assume these are all optional parameters, and they're created as standard `output` objects to the rest of the workflow?

As far as the limits on JSON goes, it's a good thing we should document but agree that 1MB of JSON is quite large

laughedelic:
Hi @febuiles @jonjanego, thanks for the feedback!

> Add examples of the usage to https://github.com/actions/dependency-review-action/blob/main/docs/examples.md

Done in 75be7f0

> Add unit tests for new features.

I'm not sure how action outputs can be tested in unit tests, since they are not like outputs of a function. But I've modified the `.github/workflows/dependency-review.yml` workflow (05fcfa4) to print out the outputs with some minimal checks: `comment-content` shouldn't be empty and the rest should pass as valid JSON through `jq`. Let me know if this is good enough as a test, or if not, please, advise on how to test it better.

> Link to a sample PR in a custom repository running your version of the Action.

Here is an example PR: https://github.com/laughedelic/dependency-review-action/pull/7 which adds a dependency with a known vulnerability. And here are the outputs from the modified workflow: https://github.com/laughedelic/dependency-review-action/actions/runs/8147727423/job/22269075026#step:5:7

> I assume these are all optional parameters, and they're created as standard output objects to the rest of the workflow?

Yes, these are standard action outputs, like the `comment-content` which was already there.

> As far as the limits on JSON goes, it's a good thing we should document but agree that 1MB of JSON is quite large

I also added this to the docs in 05fcfa4.

jonjanego:
Thank you for the contributions @laughedelic . We'll take a look at it sometime in the next couple of weeks and let you know any questions that we may have!

elireisman:
👋 hello - small update on this.

First - I like the change, thanks for contributing!

Second, I am running some tests on this Action and branch using a test repo, and will get back to you soon with feedback or a PR approval. I appreciate your patience in the meantime 🙇 

Will post an update ASAP

laughedelic:
Hi Eli! Thanks for the heads up, and no worries, take your time 🙂

src/main.ts
```diff
@@ -130,18 +130,25 @@ async function run(): Promise<void> {
     }
 
     if (config.vulnerability_check) {
+      core.setOutput('vulnerable-changes', JSON.stringify(vulnerableChanges))
```
febuiles, line 133: Is `stringify` what we want to be using if we expect people to consume the JSON?
```
vulnName = "Security Advisory for \"Log4Shell\""
=> 'Security Advisory for "Log4Shell"'
JSON.stringify(vulnName)
=> '"Security Advisory for \\"Log4Shell\\""' 
```

When I try to use it in `sh`:
```
~/w
$ cat > foo.json
[{ "name": '"Security Advisory for \\"Log4Shell\\""' }]
^D
~/w
$ cat foo.json | jq .
parse error: Invalid literal at line 1, column 13
```

If we plan to go down this route, let's move the "stringification" of output to an external function that includes unit tests.

.github/workflows/dependency-review.yml
```diff
@@ -10,5 +10,39 @@ jobs:
...
+        id: review
+
+      - name: 'Check comment-content output'
+        if: always()
+        env:
+          COMMENT: ${{ steps.review.outputs.comment-content }}
+        run: |
+          test -n "$COMMENT"
+          echo "$COMMENT"
+
```
elireisman, line 25: ☝️ it feels odd that the `comment-content` is HTML and the rest is JSON. Is this output needed. If so, should we present it in a more convenient form for use as an output?

.github/workflows/dependency-review.yml
```diff
@@ -10,5 +10,39 @@ jobs:
...
+        if: always()
+        env:
+          LICENSE_CHANGES: ${{ steps.review.outputs.invalid-license-changes }}
+        run: echo "$LICENSE_CHANGES" | jq
+
+      - name: 'Check denied-changes output'
+        if: always()
+        env:
+          DENIED_CHANGES: ${{ steps.review.outputs.denied-changes }}
+        run: echo "$DENIED_CHANGES" | jq
```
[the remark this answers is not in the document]
elireisman: I've been doing some testing with this branch of the Action, using an [output setup in my test repo](https://github.com/elireisman/monorepo/blob/elireisman/test-dr-ft/.github/workflows/dependency-review.yml) similar to what you have here, and I'm seeing some [unescaped entries in the output JSON](https://github.com/elireisman/monorepo/actions/runs/8285030033/job/22671857292)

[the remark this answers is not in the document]
laughedelic: @febuiles I added it as test, not as a demo. But sure, I can remove it.

@elireisman there is an important difference between this code and your test:
- here the output data is stored in an env var, and then in the shell script it's just a reference to the var by its name, so the data is passed directly to `jq` (see example [logs here](https://github.com/laughedelic/dependency-review-action/actions/runs/8147727423/job/22269075026?pr=7#step:5:2))
- in your workflow you do a direct reference to `${{steps.dr_test.outputs.vulnerable-changes}}` which gets interpolated _before_ the shell script is evaluated (compare the [logs here](https://github.com/elireisman/monorepo/actions/runs/8271132492/job/22630211533#step:4:2))

You could probably wrap it in single quotes: `echo '${{steps.dr_test.outputs.vulnerable-changes}}'` but it might still break because of some quotes in the JSON body (e.g. `{"foo": "a'b"}` is a valid JSON, but will break that shell syntax).

I believe that the data in the output is stored correctly and doesn't need any extra escaping, but the usage need caution and the documentation for this could probably emphasize it.

src/main.ts
laughedelic: `vulnerableChanges` as well as others is a JS object parsed from a JSON API response and `stringify` here just serializes it to a string.

In your example `JSON.stringify(vulnName) => '"Security Advisory for \\"Log4Shell\\""'` shows extra quotes because it's a JS string value (`'`) with a JSON string inside (`"`) with some escaped quotes inside

In the shell test you store an invalid JSON by keeping those single quotes. Let's do this with the whole object that you use in the second part:

```
$ node
> obj = [{ name: "Security Advisory for \"Log4Shell\"" }]
[ { name: 'Security Advisory for "Log4Shell"' } ]
> JSON.stringify(obj)
'[{"name":"Security Advisory for \\"Log4Shell\\""}]'
> fs.writeFileSync('foo.json', JSON.stringify(obj))
^D

$ cat foo.json
[{"name":"Security Advisory for \"Log4Shell\""}]

$ cat foo.json | jq
[
  {
    "name": "Security Advisory for \"Log4Shell\""
  }
]

$ cat foo.json | jq '.[].name'
"Security Advisory for \"Log4Shell\""

$ cat foo.json | jq '.[].name' --raw-output
Security Advisory for "Log4Shell"
```

As you see, it produces JSON data that can be parsed again.

docs/examples.md
```diff
@@ -190,12 +191,20 @@ jobs:
           deny-licenses: LGPL-2.0, BSD-2-Clause
       - name: 'Report'
         # make sure this step runs even if the previous failed
-        if: ${{ failure() && steps.review.conclusion == 'failure' }}
+        if: always()
```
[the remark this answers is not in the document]
laughedelic: I reverted it to the original condition in bc6a1f0dd48b6d6a9e69cae0b9499cbe36e099e8

docs/examples.md
```diff
@@ -190,12 +191,20 @@ jobs:
...
         env:
-          comment: ${{ steps.review.outputs.comment-content }}
-        run: |
-          echo "$comment"  # do something with the comment
+          COMMENT: ${{ steps.review.outputs.comment-content }}
+        run: | # do something with the comment:
+          echo "$COMMENT"
+      - name: 'List vulnerable dependencies'
+        # make sure this step runs even if the previous failed
+        if: always()
```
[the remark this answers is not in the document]
laughedelic: Done in bc6a1f0dd48b6d6a9e69cae0b9499cbe36e099e8

.github/workflows/dependency-review.yml
laughedelic: I'm not sure what you're suggesting. This output was added in https://github.com/actions/dependency-review-action/pull/698 and the use case was to post it in more places that just the PR comment. 

This change here is just for consistency with the rest of outputs. But if suggestion in https://github.com/actions/dependency-review-action/pull/707/files#r1524299054 is to remove this, this will be also gone.

febuiles: The HTML output is used for different purposes and we should be kept until we refactor how the summaries are generated, let's leave this here!

src/main.ts
febuiles: @laughedelic thanks for the explanation 🙇 

Do you see a way in which we could escape these results so people reading directly from the object's property don't get bitten by cases like `{"foo": "a'b"}` as you pointed out above?

.github/workflows/dependency-review.yml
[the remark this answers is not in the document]
elireisman: Thanks - can confirm, single quoted output still broke due to unescaped JSON values 👍 

I'm a bit concerned that the output is a bit delicate (or at least as you say requires a mention or example in the documentation?) if it requires a particular process to pass it along safely?

That said, this isn't my area of expertise 😆 so I'll defer to other reviewers if this is worth addressing or not. Thanks for the reply!

src/main.ts
elireisman: ☝️ came to the same conclusion in a 🧵 above. feels too delicate in the current form to me, but I'm not super familiar with TS so I'll keep an open mind here 😂

laughedelic:
> I'm a bit concerned that the output is a bit delicate (or at least as you say requires a mention or example in the documentation?) if it requires a particular process to pass it along safely?
> 
> I would assume the right play is to ensure the values in the JSON are escaped regardless of how it is passed around or printed, so the `output` itself is ready to use for downstream callers?

> Do you see a way in which we could escape these results so people reading directly from the object's property don't get bitten by cases like `{"foo": "a'b"}` as you pointed out above?

@febuiles @elireisman I understand your concerns, the output is "delicate" indeed. But I want to make it clear that it's not specific to this action, JS/TS or JSON data, it is about the way GitHub actions work in general: how outputs are passed around and how they get interpolated in run-steps (shell-scripts). 

Even the existing `comment-content` output which returns simple text or HTML needs such delicate treatment (especially if it is used in a shell-script), it can also contain quotes or any other characters that may break Bash syntax or even get accidentally interpreted (`$`). This is true for any "large" or multiline output data.

I think it's also a bit misleading that the only examples I'm providing here are just reading outputs in shell-script steps. My real use case is to pass them _as inputs_ to another TypeScript action which will parse JSON and do something with the data.


So I see these alternatives here:
- the approach of this PR: to return data in the output directly and document the way it can be used, maybe warn how it shouldn't be used
- an alternative is to write all data on disk in some predictable location:
  - there can be a default (outside of the working dir) and some ways to override it, e.g. with an env var or an action input
  - it can be an random place in temp + an output that provides that path. But this approach is bad for potential caching, it's better to provide a location that consumers can control.
- another approach would be to set an env var with the data directly (without outputs), but I think it's weird for arbitrary JSON data and potentially "pollutes" environment for the subsequent steps


Personally, I think that the first approach is fine since the "usage hygiene" is not specific to this action. The second approach is a bit more "heavyweight", but also works and is hard to "misuse". 

Let me know what you think and how we should proceed with this

febuiles:
@laughedelic Thanks for your patience, and for taking the time to write all of this down. The approach in this PR makes more sense than the listed alternatives, and I'm happy to move forward with it. Please revert any unnecessary changes to `.github/workflows/dependency-review.yml` (I'm thinking about the "tests") and I'll try to get this released by Wednesday.

If string sanitizing is indeed a problem, users will complain and we can learn from actual examples instead of coming up with `bash` one liners :)

laughedelic:
hey @febuiles, sounds good to me 🤝  

I reverted the changes to the workflow and added a note in the readme to explain the usage caveat, but let me know if you would prefer it in some other form

febuiles (approved):

jonjanego:
Thank you again for the contribution, @laughedelic !

laughedelic:
My pleasure! Thanks for the thorough review and a speedy release! 🚀

PR #709, opened 2024-03-12
Add support for calculating OpenSSF Scorecards

jhutchings1:
This PR adds support for pulling OpenSSF scorecards in dependency review. You can see what this looks like in action here: https://github.com/jhutchings1/dependency-review-tester/actions/runs/8126191241

Screenshot 2024-03-12 at 2 37 04 PM

cc: @actions/advanced-security-dependency-graph


elireisman (approved):
Cool feature add! 🎉

The change LGTM to me: I see tests, docs updates etc. and the code all makes sense to my eyes.

I'll stamp, with the caveat that before we ship, you may want _one_ other person on the team who has more time in maintaining this Action (and TS projects in general!) to have a peek.

Your call 👍 thanks for the contribution! 🍻

febuiles (changes requested):
By default the action does not show the HTML summary, so I [gave this a try](https://github.com/future-funk/automatic-funicular/actions/runs/8264938680/job/22609556223?pr=2) and was very happy to see a text-only report too!

There's something going on with the printing though, notice the `[object Object]` entries in the scorecard.

jhutchings1:
@febuiles Nice catch. I just pushed a fix to that one.

febuiles (approved):
Looks much nicer, thank you! https://github.com/future-funk/automatic-funicular/actions/runs/8268691764/job/22622190735?pr=3

febuiles:
@jhutchings1 Feel free to cut a new release after merging ([instructions here](https://github.com/actions/dependency-review-action/blob/main/CONTRIBUTING.md#cutting-a-new-release)). If you don't want to do this, I'll be pushing a new release on Mar 21 (next Wed).

elireisman:
🤔 not a blocker - but a few questions occurs to me, if you have a sense or some data from the repo testing you did:

1. For a sufficiently large DR changeset, do we see the scorecard fetch operations adding substantial running time to the Action?
2. Will folks using the DR API but accustomed to the DR Action wonder where the scorecards are in API responses?

I like that the fetch step is `async` and fans out nicely, hopefully this is not a problem in the 90% case 👍 but I wonder, if we see noticeable increases to DR Action running times (or API end user confusion) should we rescope this to track the scorecard data in the back end and serve it with the API responses instead?

jonjanego:
> @jhutchings1 Feel free to cut a new release after merging ([instructions here](https://github.com/actions/dependency-review-action/blob/main/CONTRIBUTING.md#cutting-a-new-release)). If you don't want to do this, I'll be pushing a new release on Mar 21 (next Wed).

Let's roll this into the release next week

jhutchings1:
> 1. For a sufficiently large DR changeset, do we see the scorecard fetch operations adding substantial running time to the Action?

Right now the performance has been quite good. For a changeset around 100 items long, the DR step is taking just 23s (so it's < 230ms/call). I doubt it'll get prohibitively long, even in very large PRs. 

> Will folks using the DR API but accustomed to the DR Action wonder where the scorecards are in API responses?

I'm not sure. This was a pretty small lift if somebody wanted to add it to their own API-based implementation though. 

> should we rescope this to track the scorecard data in the back end and serve it with the API responses instead?

I'd love to have this data on the backend service across the dependency graph, but that's a bigger lift and I was hoping to get this one out as a quick win. We can see how customers respond to this and prioritize the bigger lift based on that.

febuiles:
> If we see noticeable increases to DR Action running times (or API end user confusion) should we rescope this to track the scorecard data in the back end and serve it with the API responses instead?

If we see timing become a problem, let's write something up. We currently don't have the bandwidth to make the required changes in the backend, so I'm good with rolling with the change as it is.

Issue #670, opened 2024-01-22
Properly resolve licenses with "OR" expressions


msalib:
Package | Version | License | Issue Type
-- | -- | -- | --
ryu | 1.0.16 | Apache-2.0 OR BSL-1.0 | Incompatible License

We reject the BSL-1.0 license, so I was confused why this Rust crate was getting rejected since it can be licensed under BSL-1.0 or Apache-2.0 which we accept.

febuiles:
The SPDX expression parser we use is brittle (see https://github.com/actions/dependency-review-action/issues/263). I think moving to something like https://www.npmjs.com/package/@onebeyond/spdx-license-satisfies would provide a better experience and maybe fix the issues with `OR`. We don't have cycles to test atm, but are happy to collaborate on community contributions.

npushkarskii:
Hey there, hope this one won't slip through the cracks and be fixed at some point :) Thank you for handling this issue.

febuiles:
@npushkarskii https://github.com/actions/dependency-review-action/pull/719 needs a few tweaks, but it fixes thisbug. 

If you want to start testing it today and help us get feedback you can add `uses: actions/dependency-review-action@change-spdx-parser` in your worfklow file until the PR and the new version (probably `v5`) are merged/released.

npushkarskii:
@febuiles only managed to see your comment now, sorry for the delay. 

Thank you for all the info, my colleagues and I will try to play with the pre-release version you mentioned; if we get to it, I'll let you know about the results in this PR! Thank you again for your help.

npushkarskii:
@febuiles Hey there, unfortunately, at our company we didn't have a chance to test a pre-release version of this action due to security policies. Sorry that I got back to you with such a delay.

Hence, I just hope that this PR will be merged anytime soon so that we'll have a new release with the fixed bug 🤞🤞 Thank you!

febuiles:
cc @jonjanego

npushkarskii:
Hey there, me again :) Any chances to see this one merged anytime soon? @febuiles 

I see that the [version 4.3.4](https://github.com/actions/dependency-review-action/releases/tag/v4.3.4) partially resolves the issue, but I'm not sure if it helped in my case 🥹

Anyway, thank you for working on that and let me know if I can be of any help (even though my ability to help appeared to be [extremely limited](https://github.com/actions/dependency-review-action/issues/670#issuecomment-2127376370))

febuiles:
[...]

Issue #635, opened 2023-12-06
deny-licenses mistakenly blocking LGPL-3.0 license


logan-porelle:
# Summary:

When I list `LGPL-2.0,LGPLLR` as my deny-license list and have `PyGitHub==2.1.1` as a third party in my requirements.txt file then the dependency-review action blocks the pull request as a incompatible issue. The problem is that PyGitHub is a LGPL-3.0, not LGPL-2.0 license.

## Replicate:

1. Have dependency-review-action setup like below:

```
name: 'Dependency Review'
on: [pull_request]

permissions:
  contents: read
  pull-requests: write

jobs:
  dependency-review:
    runs-on: ubuntu-latest
    steps:
      - name: 'Checkout Repository'
        uses: actions/checkout@v3
      - name: 'Dependency Review'
        uses: actions/dependency-review-action@v3.1.4
        with:
          comment-summary-in-pr: true
          deny-licenses: LGPL-2.0,LGPLLR
```

2. Open a pull request with a requirements.txt listing `PyGithub==2.1.1`
3. Check PR dependency-review-action failure.

## Sample Output:

![Image](https://github.com/actions/dependency-review-action/assets/4312149/2c710429-89d0-43b3-985f-993dfbd7aaab)

juxtin:
Hi @logan-porelle, I've been trying to reproduce this and I haven't been able to so far. Is it still something that you're running into?

Issue #575, opened 2023-09-27
"Invalid SPDX License" after upgrading JSTS package


mprins:
After upgrading the JSTS package from 2.10.0 to 2.11.0 we're getting the message that it has an "Invalid SPDX License" "(EDL-1.0 OR EPL-1.0)"  (the license information [has not changed in 7 years](https://github.com/bjornharrtell/jsts/commit/245a98cda273f0efb9910a12fd4080ef357cb308)

We're using the following configuration
```yaml
  dependency-review:
    name: 'Dependency Review'
    if: ${{ github.event_name == 'pull_request' }}
    runs-on: ubuntu-latest
    permissions:
      contents: read
    steps:
      - uses: actions/checkout@v4

      - uses: actions/dependency-review-action@v3
        with:
          deny-licenses: GPL-2.0+, AGPL-3.0+
```

see https://github.com/B3Partners/tailormap-viewer/actions/runs/6323648494?pr=484 for the logs/output.

Do you have any hints or ideas?

possibly related to #559

febuiles:
I've reproduced this succesfully in https://github.com/future-funk/cuddly-octo-bassoon/actions/runs/6330878481/job/17194265977?pr=1. My guess is that our SPDX parser is having issues with `OR` expressions, but I will need to take a closer look. Will report my findings back!

mprins:
thanks for looking into this.

I'm not sure what the authorative source is for license keys, but I see it is lacking from https://spdx.org/licenses/ 

The `EDL-1.0` key was rejected as described on https://wiki.spdx.org/view/Legal_Team/License_List/Licenses_Under_Consideration for being identical to BSD-3-Clause.
Still, none of these are recent changes...

febuiles:
@mprins The list of licenses for NPM, Maven and Composer was recently refreshed. I'm guessing that since this is an older package there was no license set for it before. Thanks for the EDL pointer, I'll see if we can get this updated.

febuiles:
I updated the license for `jsts@2.11.0` by [asking ClearlyDefined to harvest the package](https://clearlydefined.io/harvest). Although the invalid EDL license was removed, the resulting expression is still marked invalid by the SPDX parser. Sample run here: https://github.com/future-funk/cuddly-octo-bassoon/actions/runs/6330878481/job/17558844147

ahpook:
This is an old issue, but it looks like it's covered by the improvements in compound SPDX expression handling in #927 
[...]

Issue #263, opened 2022-09-29
[Feature Request] Support for packages with multiple licenses


jcasner:
One of the packages we're importing has multiple licenses based on dependent projects. The Github dependency-graph API returns all 3 licenses. We have each of the 3 licenses in our `allow-list`, but because this is a single string response, it's failing the dependency review. 

Response from the depency-graph API:
```json
  {
    "change_type": "added",
    "manifest": "package-lock.json",
    "ecosystem": "npm",
    "name": "@fortawesome/fontawesome-free",
    "version": "5.15.4",
    "package_url": "pkg:npm/%40fortawesome/fontawesome-free@5.15.4",
    "license": "CC-BY-4.0 AND MIT AND OFL-1.1",
    "source_repository_url": "https://github.com/FortAwesome/Font-Awesome",
    "scope": "runtime",
    "vulnerabilities": []
  },
```

**Proposal**
For a situation like this, the action would parse the license field and use the operator (AND) to check that all 3 licenses are in our allow-list. I can see the possibility of a dual-licensed package including an `OR` (e.g.: `CC-BY-4.0 OR MIT`) and so we'd want to use the operator (AND or OR) to validate against the `allow-list` or `deny-list` appropriately. 

Thanks for your consideration!

febuiles:
@jcasner Thanks for the report. This is something originally reported in https://github.com/actions/dependency-review-action/pull/131 that's actively being worked on using `spdx-satisfies` to evaluate these expressions. I'll update this issue once there are news to share!

febuiles:
The latest release (`v3`) is fully SPDX-compliant and now has support for `AND`/`OR` expressions. Sadly, your specific example is a [known bug in an upstream library](https://github.com/jslicense/spdx-satisfies.js/issues/14), so I'm not confident the newest release will allow you to specific all three licenses as listed above. The `OR` expressions should work fine.

I'm leaving the bug report open until `spdx-satisfies` has been updated.

JPLachance:
Hello!

I do have a very similar issue. We have a long `allow-list` and we are constantly facing issues like the one described here.

Do we have an ETA?

febuiles:
@JPLachance we can't proceed until https://github.com/jslicense/spdx-satisfies.js/issues/14 is fixed upstream. All ears if you have suggestions on how to improve the parsing of SPDX expressions!

prakyathr:
[...]

PR #719, opened 2024-03-21
Update SPDX Expression Parsing

febuiles:


src/config.ts
```diff
@@ -142,7 +144,9 @@ async function readConfigFile(
         ref: pieces.groups.ref
       })
     } else {
-      data = fs.readFileSync(path.resolve(filePath), 'utf-8')
+      // repo-local file
+      const fullPath = path.resolve(__dirname, filePath)
+      data = fs.readFileSync(fullPath).toString()
```
mrysav, line 149: I notice the `utf-8` was dropped here, was there a particular reason for that?

src/config.ts
```diff
@@ -153,21 +157,26 @@ async function readConfigFile(
...
-      'deny-licenses',
-      'fail-on-scopes',
-      'allow-ghsas',
-      'allow-dependencies-licenses',
-      'deny-packages',
-      'deny-groups'
-    ]
+    data = YAML.parse(configData)
+  } catch (e) {
+    throw new Error(`Can't parse YAML: ${(e as Error).message}`)
```
mrysav, line 164: Not really sure what it looks like to get an error here or what is "supposed" to happen. But in another language I would wrap the error and it [looks like that might be possible here too, if I'm reading this correctly.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error)

```suggestion
    throw new Error(`Can't parse YAML: ${(e as Error).message}`, { cause: e })
```

Although, that said, is there a reason to catch the error here at all if you're just going to re-throw it?

src/config.ts
```diff
@@ -197,7 +209,7 @@ function parseConfigFile(configData: string): ConfigurationOptionsPartial {
     }
     return data
   } catch (error) {
-    throw error
+    throw new Error(`error validating config: ${error}`)
```
mrysav, line 212:
```suggestion
    throw new Error(`error validating config: ${(error as Error).message}`)
```

Worth casting to get the message here? Not sure what the output would be if this is an object.

src/config.ts
febuiles: Fixed in 161d112, ty!

src/config.ts
febuiles: I like this `cause` thingie, TIL ty! I tested it locally and it's not available in ES6 though 😢 

I introduced this after splitting nested `try` blocks, but looking at the code I don't think we're adding anything to what Node already does so I removed it in c5c5f44.

src/config.ts
febuiles: The reason to cast is to please the TS type checker, because `error` is only defined inside the `catch` we need to give it some hints so it knows it can call `message` to get the error body. I'm avoiding

__tests__/external-config.test.ts
```diff
@@ -59,38 +55,38 @@ test('it parses options from both sources', async () => {
...
-  setInput('config-file', './__tests__/fixtures/no-licenses-config.yml')
+  setInput('config-file', '../__tests__/fixtures/no-licenses-config.yml')
   const config = await readConfig()
   expect(config.fail_on_severity).toEqual('critical')
 })
 
 test('it raises an error when given an unknown severity in an external config file', async () => {
-  setInput('config-file', './__tests__/fixtures/invalid-severity-config.yml')
+  setInput('config-file', '../__tests__/fixtures/invalid-severity-config.yml')
   await expect(readConfig()).rejects.toThrow()
```
elireisman, line 83: 🤔 do these paths require the `../__tests__/` prefix or would a `fixtures/` prefix work too?

__tests__/spdx.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    expect(spdx.satisfies(license, expr)).toBe(false)
+  })
+
+  test('returns false if no matches are found', async () => {
+    const license = 'ISC'
+    const expr = 'MIT OR GPL-2.0'
+
+    expect(spdx.satisfies(license, expr)).toBe(false)
+  })
+})
```
elireisman, line 53: 😍

src/config.ts
elireisman: ☝️ I think what Mitchell is pointing out is that we're missing the `message` call after the case there?

src/spdx.ts
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    spdxParse(license)
+    return true
+  } catch (e) {
+    // make a note in the log if the exception
+    // was caused by the expression parser, this
+    // usually means a discrepancy in ESM/CJS.
+    if (e instanceof TypeError) {
+      throw new Error('spdx: invalid expression parser')
+    }
```
elireisman, line 20: 🤔 should the error message instead be `spdx: invalid expression` or something? the parser isn't the problem here, it's the inputs right?

Also under which error conditions is it safe to swallow the error and message here and just return `false` as in line 22?

elireisman:
Couple questions, but this is looking good so far, thanks! 🍻

__tests__/external-config.test.ts
febuiles: I'm not entirely sure why (maybe `jest`?), but the path needs to be relative to `src/`, hence the change from `./` to `../`:

```
[FAIL] it accepts an external configuration filename

    Unable to fetch or parse config file: ENOENT: no such file or directory, open '/Users/febuiles/w/dr-action/src/fixtures/no-licenses-config.yml'
```

src/spdx.ts
febuiles: > should the error message instead be spdx: invalid expression or something? the parser isn't the problem here, it's the inputs right?

It's the parser in the sense that `TypeError` is what's raised when there's an error loading the parser. I have not seen these errors outside of local development, where they're common if you forget to `npm i` after bootstrapping the project.

> Also under which error conditions is it safe to swallow the error

None, if this fails we should abort the run.

__tests__/external-config.test.ts
elireisman: cc @hmaurer @bteng22 thoughts on this? if its nothing, feel free to ignore 👍

src/spdx.ts
elireisman: ☝️ same here does that sync with your understanding @hmaurer ? we just bail if we can't parse a license, even if there may be vuln pkgs etc. found in this run?

again, I have little context/depth on DR Action at present, so if this is a non-concern, feel free to ignore 👍

elireisman:
👋 I've done a pretty major overhaul of this PR, as the potential behavior changes are broad and the licenses expression checking involved a lot of mocking of the SPDX expression and validity in code where the checks are applied.

Along the way I simplified the use of the new SPDX eval library, updated code that relied on error bubble-ups to short circuit license eval to be more robust in the face of illegal inputs, and added a lot of new unit tests to illustrate how the evaluation and validity checks will behave in practice.

That said, there could be surprises here and I'm no TypeScript expert 😂 so please do review this carefully 🙇 cc @hmaurer @bteng22 @mrysav

__tests__/spdx.test.ts
```diff
@@ -0,0 +1,267 @@
...
+    }
+  ]
+
+  for (const unit of units) {
+    let got: boolean = spdx.satisfiesAny(unit.candidate, unit.licenses)
+    if (got != unit.expected) {
+      console.log(
+        `failing unit test inputs: candidate(${unit.candidate}) licenses(${unit.licenses})`
+      )
+    }
```
elireisman, line 67: ☝️ why log here? allowing expression parsing will make for some counter-intuitive matching results and I wanted the tests to illustrate this going forward, so I tried to set these up in a way that adding more trials would be easy as we address corner cases in the future.

To that end, I tried to install [jest-expect-message](https://www.npmjs.com/package/jest-expect-message) but it did not play nicely with Jest + TS, and after seeking some SME advice from the team, and noting in codesearch that I don't see prior art of this in use in similar Jest + TS projects, I gave up on that.

As-is, when these tests fail, you will see which case failed due to the conditional logging. If someone with more TS than I would like to 🍐 on using that instead, or just convert back to single test cases a la vanilla Jest, let me know 👍

package.json
```diff
@@ -41,6 +41,7 @@
     "zod": "^3.22.3"
   },
   "devDependencies": {
+    "@types/jest": "^29.5.12",
     "@types/node": "^20",
     "@types/spdx-expression-parse": "^3.0.4",
     "@types/spdx-satisfies": "^0.1.1",
```
elireisman, line 47: moved Jest back into dev dependencies where I think it belongs 👍

src/licenses.ts
```diff
@@ -87,15 +87,19 @@ export async function getInvalidLicenseChanges(
...
-          const found = allow.find(spdxExpression =>
-            spdxSatisfies(license, spdxExpression)
-          )
-          validityCache.set(license, found !== undefined)
+          if (spdx.isValid(license)) {
+            const found = spdx.satisfiesAny(license, allow)
+            validityCache.set(license, found)
+          } else {
+            invalidLicenseChanges.unresolved.push(change)
+          }
```
elireisman, line 95: ☝️ the library is very strict about invalid SPDX inputs, so instead of just bucketing these as "unresolved" based on an error bubble up, we'll now check validity specifically as a pre-step when classifying and resolving licenses 👍

src/licenses.ts
```diff
@@ -161,10 +165,11 @@ const setGHLicenses = async (changes: Change[]): Promise<Change[]> => {
 
   return Promise.all(updatedChanges)
 }
+
 // Currently Dependency Graph licenses are truncated to 255 characters
 // This possibly makes them invalid spdx ids
 const truncatedDGLicense = (license: string): boolean =>
-  license.length === 255 && !isSPDXValid(license)
+  license.length === 255 && !spdx.isValid(license)
```
elireisman, line 173: 🤔 this old codepath could be rethought - just a validity check (required anyway) on our inputs will catch these without the specific truncation checks. I didn't bother in this PR

src/spdx.ts
```diff
@@ -0,0 +1,56 @@
...
+export function satisfiesAll(
+  candidateExpr: string,
+  licenses: string[]
+): boolean {
+  try {
+    return spdxlib.satisfiesAll(candidateExpr, licenses)
+  } catch (_) {
+    return false
+  }
+}
```
elireisman, line 46: ☝️ The conditions that throw an error in the new SPDX parsing and eval are very strict, and this made testing and using the library more awkward, so I moved to just return `false` in these cases, and in cases where error bubble up was a sentinel in the logic, I moved to apply an explicit validity check on the inputs instead.

I can go into more detail if desired about why Jest didn't play nice with the error bubble up previously for those that I didn't chat with about this already, but I'm pretty happy with the change so far. Let me know if you disagree or want to sync about it though 👍

__tests__/spdx.test.ts
juxtin: Agreed, I'm going to remove that.

__tests__/spdx.test.ts
```diff
@@ -59,17 +59,14 @@ test('satisfiesAny', () => {
...
-      console.log(
-        `failing unit test inputs: candidate(${unit.candidate}) licenses(${unit.licenses})`
-      )
-    }
-    expect(got).toBe(unit.expected)
+    const got: boolean = spdx.satisfiesAny(unit.candidate, unit.licenses)
+    test(`should return ${unit.expected} for ("${unit.candidate}", "${unit.licenses}")`, () => {
+      expect(got).toBe(unit.expected)
+    })
   }
```
elireisman, line 66: oh nice change! 😍

src/licenses.ts
elireisman: 🤔 due to this change, we may be able to ditch the `try...catch` blocks entirely

src/licenses.ts
```diff
@@ -87,15 +87,19 @@ export async function getInvalidLicenseChanges(
...
-            spdxSatisfies(license, spdxExpression)
-          )
-          validityCache.set(license, found === undefined)
+          if (spdx.isValid(license)) {
+            const found = spdx.satisfiesAny(license, deny)
+            validityCache.set(license, !found)
+          } else {
+            invalidLicenseChanges.unresolved.push(change)
+          }
         }
```
elireisman, line 103: ⚠️ one more note about the new uses of `satisfiesAny` in these evaluations. This should correctly resolve the problems listed in the support Issues that this PR will close out ...except for [this one](https://github.com/actions/dependency-review-action/issues/263). 

As the new test suites for SPDX expression eval show, we could use `satisfies` in place of `satisfiesAny` to make not only the incoming package license spec parse as an expression (as this PR enables) but also treat our allow/deny list entries as expressions. This would enable DR Action config to set not only single-license entries, but expressions like `MIT AND Apache-2.0` and so on.

But, to resolve the case described in the linked support Issue (incoming package includes an `AND` expression) the end user would still need to supply some duplicative entries in the expression-based allow/deny lists to ensure a good match...

Seems like closing out that last Issue should happen after this PR, and after mulling over some product-facing options for supporting these expressions better 👍

__tests__/licenses.test.ts
```diff
@@ -30,7 +29,7 @@ const rubyChange: Change = {
   name: 'actionsomething',
   version: '3.2.0',
   package_url: 'pkg:gem/actionsomething@3.2.0',
-  license: 'BSD',
+  license: 'BSD-3-Clause',
```
juxtin, line 32: Just noting this here—`BSD` is not a valid license type in earlier versions of this action either, but tests used to bypass that check.

src/licenses.ts
juxtin: Thanks @elireisman, the description for this PR indicated that it closes that issue but I was able to confirm that it does not. I removed that line from the description :(

src/licenses.ts
juxtin: Yeah, that's a reasonable take. I'll also leave it for now.

juxtin:
From a code review/testing perspective, I'm pretty happy with this right now. My plan is to do a little bit of manual testing to validate that the linked issues really are resolved, and then I'll get this approved, merged, and released.

juxtin:
> From a code review/testing perspective, I'm pretty happy with this right now. My plan is to do a little bit of manual testing to validate that the linked issues really are resolved, and then I'll get this approved, merged, and released.

I'm glad I didn't rush through this, because this PR doesn't resolve _any_ of the issues that were linked above. That's not to say it's a bad PR or that we shouldn't merge it, but it doesn't address the fundamental causes of the behaviors that people have run into.

I'll break it down issue by issue:
* #670 
    * problem: `Apache-2.0 OR BSL-1.0` should be allowed when only `Apache-2.0` is denied. One of its acceptable licenses is implicitly allowed, so it should be fine.
    * reality: as of this PR, we use `spdx.satisfiesAny('Apache-2.0 OR BSL-1.0', 'Apache-2.0')` to evaluate this. It returns `true`, because 'Apache-2.0 OR BSL-1.0' _does_ satisfy 'Apache-2.0'. This just isn't the semantics that we want.
* #575 
    * problem: `EPL-1.0 OR NOASSERTION OR (EPL-1.0 AND NOASSERTION)` is considered an `Invalid SPDX License`.
    * reality: as of this PR, `NOASSERTION` is _still_ considered an invalid license.
* #635 
    * problem: `LGPL-3.0` license is denied when the deny list only contains `LGPL-2.0, LGPLLR`
    * reality: for reasons I haven't totally figured out yet, this is still the behavior as of this PR. For what it's worth, internally we appear to have a license of `GPL-3.0 AND GPL-3.0-only AND GPL-3.0-or-later AND LGPL-3.0-only` for pygithub, but it is displayed in the action summary as `LGPL-3.0`.

juxtin (approved):
Despite the fact that this doesn't fix any outstanding bugs, I do still think it moves us in the right direction. The remaining bugs will have to be addressed by other work in the future.

PR #720, opened 2024-03-21
Remove Ruby from devcontainer

jhutchings1:
Our devcontainer configuration currently includes Ruby, which slows down Codespace creation and isn't necessary. This PR removes that from the configuration

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,8 +2,5 @@
   "name": "Dependency Review Action",
   "image": "mcr.microsoft.com/devcontainers/typescript-node:18",
   "postCreateCommand": "npm install",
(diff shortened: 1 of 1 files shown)


elireisman:
🤔 No strong opinions here, but I think it is used for dev debugging/triage scripts [like this](https://github.com/actions/dependency-review-action/blob/0fa40c3c10055986a88de3baa0d6ec17c5a894b3/scripts/scan_pr) cc @febuiles @hmaurer

febuiles:
@jhutchings1 we shouldn't remove because of`script/scan_pr` as Eli mentioned. 

The script is fairly simple though, if this becomes annoying I'd recommend asking Copilot to rewrite `scan_pr` in JS to remove the Ruby dependency.

PR #722, opened 2024-03-22
Revert default values in action.yml to fix external configs

### LINKED ISSUE #706: Some Settings In External Configuration Overridden By Inline Defaults
## The Problem

I discovered that the `warn-only` parameter is not settable in an external configuration file by first [seeing it get ignored when set only there](https://github.com/AlexWilson-GIS/test-dependency-review/pull/11), then seeing it honored when [it was set inline](https://github.com/AlexWilson-GIS/test-dependency-review/pull/13). I believe that this is the source of the issue:

The default value for `warn-only` is set here:
https://github.com/actions/dependency-review-action/blob/40eb2b8b00a215a5a327a2de3fdf3d6ab861c1fb/action.yml#L64-L67

This means that this parameter, along with every other parameter that has a default value defined in `action.yml`, will have an inline setting when the code is merging configurations. However, the way that they are merged is problematic:
https://github.com/actions/dependency-review-action/blob/40eb2b8b00a215a5a327a2de3fdf3d6ab861c1fb/src/config.ts#L19-L22

As documented in [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals), whenever two objects are merged using the spread operator, fields in a later object will override fields from an earlier object with the same key. This makes `warn-only` and other settings with default values un-configurable from an external file. This is a problem if you want to distribute one configuration file across PR checks in many repositories, including some owned by others that are subscribing to your settings.

## Suggested Fixes

1. Flip the order in which the settings are merged with the spread operator so that parameters set in the external configuration file override inline settings.
2. Add another boolean parameter that dictates if the inline settings override the external settings, with a default value of `false` in order to fix the bug.
3. Simply remove the default values from all of the parameters except one. They are all denoted as not required in `action.yml`:
https://github.com/actions/dependency-review-action/blob/40eb2b8b00a215a5a327a2de3fdf3d6ab861c1fb/action.yml#L7-L10
https://github.com/actions/dependency-review-action/blob/40eb2b8b00a215a5a327a2de3fdf3d6ab861c1fb/action.yml#L56-L67
...and they are all pulled into the action using functions that denote they are optional:
https://github.com/actions/dependency-review-action/blob/40eb2b8b00a215a5a327a2de3fdf3d6ab861c1fb/src/config.ts#L44-L50
The one exception to this is `repo-token`:
https://github.com/actions/dependency-review-action/blob/40eb2b8b00a215a5a327a2de3fdf3d6ab861c1fb/src/dependency-graph.ts#L13
...however, tokens shouldn't be hardcoded into config files anyway, so that's not an issue like the others.
4. Move the setting of each default value into the underlying code after checking to see if the parameter in question is not set in the inline configuration or an external settings file. This is the only suggested solution that isn't a breaking change, but it also makes the code more messy, since there are now three places that parameters could be set.

febuiles:
Closes #706.

### Context

Dependency Review Action supports configuration by three different methods:

* Inline configuration (writing your config options in your YML)
* In-repo config file (your config lives in a YML file inside your repo)
* External config (your config file lives in a repo that you have access to)

In order to handle these three cases, we don't allow adding `default` fields to the project's `action.yml`.

### Description

A few of the recently introduced options have modified the default values in `action.yml`, ignoring the comment at the top, and breaking integrations with external configuration files. 

This PR rewrittes the comment at the top (making it bigger should call more attention to it), and I removed the `default` field from the config options that had it [^0].

A proper solution given more time would be to introduce a linter that checks for non-token `default:` entries in `action.yml` to block merging the changes. Issue created here: https://github.com/actions/dependency-review-action/issues/723

[^0]: The repo token is an exception to this, since we need to have it set by the time the Zod parsers are loaded.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,5 +1,13 @@
-# Avoid using default values for options here since they will
-# end up overriding external configurations.
+# IMPORTANT
+#
+# Avoid setting default values for configuration options in
+# this file, they will overwrite external configurations.
+#
+# If you are trying to find out the default value for a config
+# option please take a look at the README or src/schemas.ts.
+#
+# If you are adding an option, make sure the Zod definition
+# contains a default value.
 name: 'Dependency Review'
 description: 'Prevent the introduction of dependencies with known vulnerabilities'
 author: 'GitHub'
@@ -56,23 +64,18 @@ inputs:
   retry-on-snapshot-warnings:
     description: Whether to retry on snapshot warnings
     required: false
-    default: false
   retry-on-snapshot-warnings-timeout:
     description: Number of seconds to wait before stopping snapshot retries.
     required: false
-    default: 120
   warn-only:
     description: When set to `true` this action will always complete with success, overriding the `fail-on-severity` parameter.
     required: false
-    default: false
   show-openssf-scorecard:
     description: Show a summary of the OpenSSF Scorecard scores.
     required: false
-    default: true
   warn-on-openssf-scorecard-level:
     description: Numeric threshold for the OpenSSF Scorecard score. If the score is below this threshold, the action will warn you.
     required: false
-    default: 3
 outputs:
   comment-content:
     description: Prepared dependency report comment


juxtin:
Do we set default values for these options in code?

mrysav (approved):
Giving this a 👍 to fix the current situation although we'll follow up outside of the PR for longer term steps.

PR #725, opened 2024-03-22
Bug fixes to #718

### LINKED ISSUE #718: Invalid URL for OpenSSF Scorecard Package
In the latest update to this action, the [OpenSSF Scorecard table](https://github.com/dupuy/reliabot/actions/runs/8379385585#summary-22946157404) generated by the dependency review action for the dependabot PR in my project's repo to update the version of that action, I get incorrect labels with invalid URLs:

[actions/actions/dependency-review-action](https://https//github.com/actions/dependency-review-action)
```
[actions/actions/dependency-review-action](https://https//github.com/actions/dependency-review-action)
```

As you can see, `actions/` is repeated in the anchor text. I'm not sure how or where in the code this is happening.

I suspect the invalid URL is due to a faulty URL in the data (so, not necessarily a bug in the code), but I'm not sure where in the data this mistake (missing `:` after `https` in URL) was made.

However, the [code in `src/scorecard.ts`](https://github.com/actions/dependency-review-action/blob/0fa40c3c/src/scorecard.ts#L16-L18) is already messing with the URL:
```typescript
   //If the repository_url includes the protocol, remove it
    if (repositoryUrl?.startsWith('https://')) {
      repositoryUrl = repositoryUrl.replace('https://', '')
    }
```

The logic here could be enhanced to handle this not-uncommon error (as well as the possibility of an insecure `http:` prefix). This might best be done by using a regex like `https?:?//` for the check and replace.

jhutchings1:
Fixes #718 
This pull request addresses the handling of GitHub Actions in the `scorecard.ts` and `scorecard.test.ts` files. The changes include the addition of a special case for GitHub Actions in the `getScorecardLevels` function, modification of the `apiRoot` URL, and the addition of new tests to verify these changes.


src/scorecard.ts
```diff
@@ -17,8 +17,16 @@ export async function getScorecardLevels(
...
     }
 
+    // Handle the special case for GitHub Actions, where the repository URL is null
+    if (ecosystem === 'actions') {
+      // The package name for GitHub Actions in the API is in the format `owner/repo/`, so we can use that to get the repository URL
+      // If the package name has more than 2 slashes, it's referencing a sub-action, and we need to strip the last part out
+      const parts = packageName.split('/')
+      repositoryUrl = `github.com/${parts[0]}/${parts[1]}` // e.g. github.com/actions/checkout
+    }
+
```
elireisman, line 27: 🤔 nit: would be good to explicitly check this output in the new test case?

elireisman (approved):
Nice change! One callout but happy to restamp after an update if you agree

src/scorecard.ts
jhutchings1: Hmm, the repositoryUrl isn't exposed outside of the function. Thoughts on how to bootstrap that into the tests? Was trying to avoid side effects on the Change object by using a local variable.

elireisman: Sorry late getting back to this. May the easiest play is to factor out 21-26 into a function that accepts a `Change` and outputs a `repositoryUrl` (or not) that line 29 can check, then test that in isolation with some Actions-flavored `Change`s?

Again not the end of the world, since it's already landed 👍

Issue #705, opened 2024-02-28
Package Deny List Ignores Version


AlexWilson-GIS:
https://github.com/actions/dependency-review-action/blob/40eb2b8b00a215a5a327a2de3fdf3d6ab861c1fb/src/deny.ts#L14
https://github.com/actions/dependency-review-action/blob/40eb2b8b00a215a5a327a2de3fdf3d6ab861c1fb/src/deny.ts#L18

After losing hours trying to figure out if there's some syntax error in my configuration file causing [1 package not on my deny list to be denied over and over](https://github.com/AlexWilson-GIS/test-dependency-review/pull/3), I crack open the source code to discover that the action doesn't actually pay attention to versions in the purl notation. This is quite surprising, as the use case in my test repo of denying older versions of a package where we specifically want to avoid regressions is a huge reason to use this action. It's also strange that this isn't documented in the README, which simply states that you can specify packages to be denied using the purl format, and the purl format includes an optional version string. I know that there is a draft spec for annotating version ranges in purl format, but until that is done, can this action please be enhanced to at least check the hardcoded version if one exists? Only being able to deny entire packages, regardless of version, is way too broad for us to make use of in many situations. If someone wants to ban an entire package, regardless of version, they could simply not provide a version string, and the current behavior could handle that situation.

febuiles:
@AlexWilson-GIS thanks for bringing this to our attention. I was under the impression that PURL denylists would also take versions into account too, and wrote some of the docs with that assumption in mind!

We'll update the issue once we have some free cycles to work on this, but if you want to PR the change and need some support feel free to reach out, I'd be more than happy to help get this fixed.

juxtin:
So I see three options here, each with pros and cons:
1. Do nothing. I think this is probably **not** acceptable for the reasons that @AlexWilson-GIS pointed out. It sounds perfectly reasonable to me to allow people to deny some versions of a package without denying all of them.
2. Match on the exact version provided in the deny list, if provided. If there's no version provided, treat it as a wildcard and deny all versions. This would be pretty simple to implement.
3. One way or another, make it possible to deny a specified range of versions. This would be easy if the purl spec supported version ranges, but it does not.

For an initial fix, I'll give my vote to option 2. I would certainly be open to supporting ranges somehow in the future, but I don't think the team has the bandwidth for that at the moment.

PR #733, opened 2024-03-27
deny-packages configuration option can deny specified version or all packages

febuiles:
### Context
- Fixes https://github.com/actions/dependency-review-action/issues/705

### What are you trying to accomplish?

https://github.com/actions/dependency-review-action/issues/705 surfaces an issue with `deny-packages` configuration option. This PR updates the option to match on exact versions, or if no version is provided to treat it as a wildcard and deny all packages regardless of version.

The PR also includes some light refactoring of the tests and code.

### 2024-04-16 Updated validation screenshots
Using this PR for testing https://github.com/bteng22/monorepo/pull/1

Denying packages with exact version
denies-exact-version

Denying all versions of package 
denies-all-versions

Denying all packages from group
denies-matching-group

Incorrectly formatted PackageURLs errors are thrown
forcing-error


src/deny.ts
```diff
@@ -3,36 +3,37 @@ import * as core from '@actions/core'
...
     }
 
-    if (deniedGroups) {
-      for (const denied of deniedGroups) {
-        if (packageUrl.startsWith(denied.toLowerCase())) {
-          changesDenied.push(change)
-          failed = true
-        }
+    for (const denied of deniedGroups) {
+      if (packageName.startsWith(denied.toLowerCase())) {
```
juxtin, line 29: The use of `startsWith` here makes me a little nervous. For example, if someone wanted to deny [pkg:npm/j](https://www.npmjs.com/package/j), would that also end up denying `pkg:npm/jest`?

src/deny.ts
```diff
@@ -3,36 +3,37 @@ import * as core from '@actions/core'
...
   const changesDenied: Change[] = []
 
-  let failed = false
+  let hasDeniedPackage = false
   for (const change of changes) {
-    change.name = change.name.toLowerCase()
-    const packageUrl = change.package_url.toLowerCase().split('@')[0]
+    const [packageName, packageVersion] = change.package_url
+      .toLowerCase()
+      .split('@')
```
juxtin, line 15: Getting the package version from a purl is a little more complicated than this. The [spec](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst) allows for additional "qualifiers" after the version, e.g.:

```
pkg:rpm/fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25
```

We may not need to worry about that in the deny list, but I think we should be prepared for those extra segments in the change list because they might get returned from snapshots.

juxtin (changes requested):
Left a couple of notes, but this is looking pretty good overall!

src/deny.ts
bteng22: Good catch! Matching on substrings like this could cause issues like in your example. I'm thinking to deny against exact namespace matches so there's no ambiguity.

bteng22:
hey @juxtin this one's ready for re-review whenever you get the chance. I used the packageurl-js library to parse out our PURLs to avoid any ambiguity between matching the changed packages and the configuration options. I updated the screenshots in the original post for validation 👍  

The gotcha' here is that our `deny-groups` option isn't a correctly formatted PURL. It's missing a package name field which makes sense. Wrote a hacky workaround for the time being https://github.com/actions/dependency-review-action/pull/733/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R73 
and opened an issue on the library https://github.com/package-url/packageurl-js/issues/67. Open to other ideas here

juxtin (approved):
👍🏽 nice!

PR #735, opened 2024-03-28
Readme action variable name for scorecard is wrong

lukehinds:
The actual name from action.yml is `show-openssf-scorecard` and not `show-openssf-scorecard-levels` as shown in the README.md


febuiles:
Great catch, thank you very much!

Issue #759, opened 2024-04-29
Configuring allow-dependencies-licenses fails the action


jdavis-etdx:
The following change made in #753 is now causing using the `allow-dependencies-licenses` configuration to fail the action with the following error:  `Error: Invalid purl(s) in allow-dependencies-licenses:...` when configured. This logic is indicating that if the the error property for an object in the list is null, add it to the `invalid_purls` list in which an an exception if this list has at at least one item.  The correct logic should be to only add an item to the list if the error property contains a value.

https://github.com/actions/dependency-review-action/blob/e58c696e52cac8e62d61cc21fda89565d71505d7/src/config.ts#L235-L243

juxtin:
Thanks for the report, @jdavis-etdx! I see you also opened and closed a PR for this, but I've gone ahead and created another solution in https://github.com/actions/dependency-review-action/pull/761.

jdavis-etdx:
Excellent! I am just happy to get it resolved. Thanks @juxtin

juxtin:
@jdavis-etdx v4.3.2 has been released and the v4 tag has been updated, so you shouldn't see this issue again.

PR #761, opened 2024-04-29
Fix package-url parsing for allow-dependencies-licenses

juxtin:
Closes https://github.com/actions/dependency-review-action/issues/759

Unfortunately we didn't have tests for this config parameter, and it was using a sort of odd validation flow. I've added a few tests and brought it up to spec with the other package-url types for the most part, but I decided to leave it as a string for now to avoid having to significantly rewrite the actual license checking logic.


src/schemas.ts
```diff
@@ -46,6 +46,22 @@ const PackageURLWithNamespace = z
     }
   })
 
+const PackageURLString = z.string().superRefine((value, context) => {
```
bteng22, line 49: 👍  Nice I really like zod's superRefine. Hopefully we can slowly tease out more of the validation logic out of the config.ts and into their own schema's here.

bteng22 (approved):
LGTM!

Issue #764, opened 2024-05-02
Adding a license in 'allow-dependencies-licenses' does not prevent it from being populated in "invalid-license-changes"


sreya:
My expectation is that adding a license to `allow-dependencies-licenses` should prevent it from being populated in the `invalid-license-changes` output however that doesn't seem to be the case with the action below. Is this a bug or are my expectations for the output incorrect? If I misunderstood, how do I manually approve unlicensed packages so that they do not fail the job

```yaml
  dependency-license-review:
    runs-on: ubuntu-latest
    if: github.ref != 'refs/heads/main'
    steps:
      - name: "Checkout Repository"
        uses: actions/checkout@v4
      - name: "Dependency Review"
        id: review
        # TODO: Replace this with the latest release once https://github.com/actions/dependency-review-action/pull/761 is merged.
        uses: actions/dependency-review-action@49fbbe0acb033b7824f26d00b005d7d598d76301
        with:
          allow-licenses: Apache-2.0, BSD-2-Clause, BSD-3-Clause, CC0-1.0, ISC, MIT, MIT-0, MPL-2.0
          allow-dependencies-licenses: "pkg:golang/github.com/pelletier/go-toml/v2, pkg:golang/gopkg.in/DataDog/dd-trace-go.v1@1.63.1"
          license-check: true
          vulnerability-check: false
      - name: "Report"
        # make sure this step runs even if the previous failed
        if: always()
        shell: bash
        env:
          VULNERABLE_CHANGES: ${{ steps.review.outputs.invalid-license-changes }}
        run: |
          fields=( "unlicensed" "unresolved" "forbidden" )
          for field in "${fields[@]}"; do
            # Use jq to check if the array is not empty
            if [[ $(echo "$VULNERABLE_CHANGES" | jq ".${field} | length") -ne 0 ]]; then
              echo "$VULNERABLE_CHANGES" | jq
              exit 1
            fi
          done
          echo "No incompatible licenses detected"
```

```
{
  "unlicensed": [
    {
      "change_type": "added",
      "manifest": "go.mod",
      "ecosystem": "gomod",
      "name": "gopkg.in/DataDog/dd-trace-go.v1",
      "version": "1.63.1",
      "package_url": "pkg:golang/gopkg.in/DataDog/dd-trace-go.v1@1.63.1",
      "license": null,
      "source_repository_url": null,
      "scope": "runtime",
      "vulnerabilities": []
    }
  ],
  "unresolved": [],
[...]

PR #765, opened 2024-05-02
Allow slashes in purl package names

juxtin:
See https://github.com/actions/dependency-review-action/issues/764.

You are _not supposed_ to have `/` characters in a package name, but people intuitively want to put them in there anyway and it's still fully possible to parse them. Since the goal of our purl handling here is maximum permissiveness within reason, I made a little tweak to the parser so that we can handle this case.


mrysav (approved):
I'm kind of uncomfortable with this as it actually might take a stance on the ambiguity eg. is the namespace `gopkg.in/DataDog` or just `gopkg.in`? 

However this appears to be an unresolved point of discussion in the PURL community: https://github.com/package-url/purl-spec/issues/63

So I think increasing permissiveness on our end is fine.

juxtin:
@mrysav that's a good callout, and we may have to revisit that later if the purl-spec folks ever work this out.

PR #766, opened 2024-05-03
fix: getRefs function to handle merge_group events

### LINKED ISSUE #456: Support merge_group trigger
When using this action with the new merge queue feature, I get the following error:
```
Error: Both a base ref and head ref must be provided, either via the `base_ref`/`head_ref` config options, or by running a `pull_request`/`pull_request_target` workflow.
```

It would be awesome, if this action can support the `merge_group` trigger out of the box.

louis-bompart:
Fixes #456

Essentially, put https://github.com/actions/dependency-review-action/issues/456#issuecomment-1537840047 in the code.

To the best of my knowledge, this should work, because a `merge_group` event is a [PullRequestEvent](https://docs.github.com/en/rest/using-the-rest-api/github-event-types?apiVersion=2022-11-28#pullrequestevent) I think.


louis-bompart:
👋 @elireisman, sorry for the poke; I was hoping to get a review on this 😶. I think it's a quick win ;) 
Feel free to redirect me to someone else if I'm barking at the wrong tree 😄 
Thanks!

elireisman (approved):
This LGTM - and sorry I didn't see this alt. PR sooner, I think I last interacted on the other that was closed out?

I'll also ping our new FR for a second pair of 👀 before we land it cc @pcarlisle 

Thanks again for the contribution! ❤️

louis-bompart:
Thanks 🙌
I saw 47a6b35 :x: on https://github.com/actions/dependency-review-action/actions/runs/8933751781/job/24923072522
I'm assuming this was caused by my branch being a few commits behind main, so I rebased my branch onto main with 92ca1f7

elireisman:
👋 I think you need to update the PR one more time prior to re-review (see the `check-dist` entry)

louis-bompart:
> 👋 I think you need to update the PR one more time prior to re-review (see the `check-dist` entry)

My bad! I missed the `npm run package` 😅 before committing. Should be fixed in e7e99e1

nachtgold:
Hi @elireisman do you know another possible reviewer than @pcarlisle?

elireisman:
👋 @nachtgold @louis-bompart apologies - looks like there wasn't time for anyone to circle back to this yet.

Looks like we need one more dist update (`npm run package`) pushed to the PR to get it updated before we can move forward. Unfortunately, if I check out the branch and do that myself, I won't be able to review the PR either 😄

juxtin (approved):

PR #771, opened 2024-05-15
Add trusty scores

therealnb:
This is some work to include [trusty](https://stacklok.com/trusty) scores in the dependency review action.

There are some more unit tests and this has been manually tested with https://github.com/StacklokLabs/DepRevTest/actions.

Many thanks.

CC @lukehinds


jonjanego:
Hi @lukehinds and @therealnb !

Some initial comments and observations:

* In the results, what is the "+ / -" column indicating?  It seems to be including two pieces of information, a "+" or "-", then an overall stoplight measure for the package.  I think this should be broken into two separate columns, although i'm not sure what the "+" or "-" indicates?
* This PR also includes a change to [dependency-review.yml](https://github.com/actions/dependency-review-action/blob/main/.github/workflows/dependency-review.yml) which is the workflow definition for the repository that runs dependency review on itself. It's included a change to run it against your action - which we won't be able to merge.

I'll make some other suggestions and feedback inline in the PR

README.md
```diff
@@ -85,6 +85,14 @@ Configure this action by either inlining these options in your workflow file, or
...
 | `warn-on-openssf-scorecard-level`      | When `show-openssf-scorecard-levels` is set to `true`, this option lets you configure the threshold for when a score is considered too low and gets a :warning: warning in the CI.                         | Any positive integer                                                                                         | 3             |
+| `trusty-scores`| `trusty-scores` is a boolean to enable or disable Trusty scores | `true`, `false` | `true` |
+| `trusty-retries`| `trusty-retries` specifies the number of retries the action uses to fetch from the trusty API. | Any positive integer | 3 |
+| `trusty-show`| `trusty-show` is the minimum score package to show. | Any positive integer | 7 |
+| `trusty-warn`| `trusty-warn` is the minimum score before a wqarning is shown. | Any positive integer | 5 |
+| `trusty-fail`| `trusty-fail` is the minimum score before a failure is shown and the action is marked as failed. | Any positive integer | 1 |
+| `trusty-api`| `trusty-api` is the base url for API calls. There should be no need to change this. | string | 'https://api.trustypkg.dev' |
```
jonjanego, line 93:
```suggestion

```

If this isn't needed  / expected user input we should drop it from the README

README.md
```diff
@@ -85,6 +85,14 @@ Configure this action by either inlining these options in your workflow file, or
...
+| `trusty-ui`| `trusty-ui` is the base url for the Trusty UI. There should be no need to change this. | string | 'https://trustypkg.dev' |
```
jonjanego, line 94:
```suggestion

```
If this isn't needed / expected user input we should drop it from the README

README.md
```diff
@@ -85,6 +85,14 @@ Configure this action by either inlining these options in your workflow file, or
...
 | `warn-on-openssf-scorecard-level`      | When `show-openssf-scorecard-levels` is set to `true`, this option lets you configure the threshold for when a score is considered too low and gets a :warning: warning in the CI.                         | Any positive integer                                                                                         | 3             |
+| `trusty-scores`| `trusty-scores` is a boolean to enable or disable Trusty scores | `true`, `false` | `true` |
+| `trusty-retries`| `trusty-retries` specifies the number of retries the action uses to fetch from the trusty API. | Any positive integer | 3 |
+| `trusty-show`| `trusty-show` is the minimum score package to show. | Any positive integer | 7 |
+| `trusty-warn`| `trusty-warn` is the minimum score before a wqarning is shown. | Any positive integer | 5 |
```
jonjanego, line 91:
```suggestion
| `trusty-warn`| `trusty-warn` is the minimum score before a warning is shown. | Any positive integer | 5 |
```

README.md
```diff
@@ -85,6 +85,14 @@ Configure this action by either inlining these options in your workflow file, or
...
 | `show-openssf-scorecard-levels`        | When set to `true`, the action will output information about all the known OpenSSF Scorecard scores for the dependencies changed in this pull request.                                                     | `true`, `false`                                                                                              | `true`        |
 | `warn-on-openssf-scorecard-level`      | When `show-openssf-scorecard-levels` is set to `true`, this option lets you configure the threshold for when a score is considered too low and gets a :warning: warning in the CI.                         | Any positive integer                                                                                         | 3             |
+| `trusty-scores`| `trusty-scores` is a boolean to enable or disable Trusty scores | `true`, `false` | `true` |
```
jonjanego, line 88: Let's change this to `false` by default until we have a chance to test and get some user feedback.  Also in code - not sure where that is living.

.github/workflows/dependency-review.yml
```diff
@@ -11,4 +11,4 @@ jobs:
       - name: 'Checkout Repository'
         uses: actions/checkout@v4
       - name: Dependency Review
-        uses: actions/dependency-review-action@main
+        uses: StacklokLabs/trusty-dependency-review-action@add_trusty_scores
```
jonjanego, line 14:
```suggestion
        uses: actions/dependency-review-action@main
```

please do not touch this workflow file

src/schemas.ts
```diff
@@ -110,6 +130,13 @@ export const ConfigurationOptionsSchema = z
     retry_on_snapshot_warnings_timeout: z.number().default(120),
     show_openssf_scorecard: z.boolean().optional().default(true),
     warn_on_openssf_scorecard_level: z.number().default(3),
+    trusty_scores: z.boolean().optional().default(true),
```
jonjanego, line 133: lets change the default to false please

jonjanego (changes requested):
Some initial feedback; not a tech review

README.md
therealnb: Ok, makes sense. Will address this.

README.md
therealnb: Will do.

src/schemas.ts
therealnb: Will do.

therealnb:
@jonjanego thanks for the feedback.

 +/- is whether the file was added or removed. The next symbol is whether this is a good idea or not. Basically, removing any file is safe (green tick). Adding one with a score lower than the configured levels might get a warning or a cross.

I did consider only showing files that were added, but it might be useful context to show the ones that are removed too (i.e. removing a high scoring one and replacing it with a low scoring one). I am open to suggestions here.

I'll work on your other comments.

Cheers

jonjanego:
> @jonjanego thanks for the feedback.
> 
> +/- is whether the file was added or removed. The next symbol is whether this is a good idea or not. Basically, removing any file is safe (green tick). Adding one with a score lower than the configured levels might get a warning or a cross.
> 
> I did consider only showing files that were added, but it might be useful context to show the ones that are removed too (i.e. removing a high scoring one and replacing it with a low scoring one). I am open to suggestions here.
> 
> I'll work on your other comments.
> 
> Cheers

thanks for the clarification.  i think it's fine to include that, but a bit of UX feedback would be to split that information into two separate columns.  it's a cleaner view that way.

therealnb:
I just saw your comment after I committed this code
https://github.com/StacklokLabs/trusty-dependency-review-action/commit/73dc7069e288507216620f8c7fe55dc495dc637a
Which is another way to do it.

See https://github.com/StacklokLabs/DepRevTest/actions/runs/9131982560?pr=5 for an example.

If you want I can revert and we can have two columns. Let me know.

therealnb:
For information these ones

Dependency Change | Version | Score | Not Malicious | Not Deprecated | Not Archived
-- | -- | -- | -- | -- | --
❌ added bugsnagmw | 1.0.3 | 0 | ❌ | ✅ | ✅ |  
⚠️ added psycopg | 3.1.18 |   | ✅ | ✅ | ✅ | 503 Service Unavailable
⚠️ added psycopg_pool | 3.2.1 |   | ✅ | ✅ | ✅ | 200 failed
⚠️ added pyarrow | 16.0.0 |   | ✅ | ✅ | ✅ | 503 Service Unavailable
⚠️ added python_json_logger | 2.0.7 |   | ✅ | ✅ | ✅ | 200 failed
⚠️ added scikit_learn | 1.4.2 |   | ✅ | ✅ | ✅ | 200 failed
⚠️ added slack_sdk | 3.27.1 |   | ✅ | ✅ | ✅ | 200 failed

Are a bug I found in our ingestion. There is a fix for those that should be deployed next week.

jonjanego:
Hi, as discussed with @lukehinds we are going to close this PR.

PR #776, opened 2024-06-03
fix show-openssf-scorecard-levels input

ramann:
The input in the README was wrong, I have fixed it to match action.yml.


jonjanego:
wow. good catch, thanks @ramann !

jonjanego (approved):

Issue #825, opened 2024-08-30
Add option for commit status check


ebickle:
**Is your feature request related to a problem? Please describe.**
I'd like the dependency review action to have an option to create a commit status check to indicate the success or failure of the dependency review independently from the status of the workflow itself. 

This would allow the dependency review action to work similarly to CodeQL - the success of the action itself would be (optionally) independent from the "dependency failures", allowing developers to enforce that dependency review is "run" but still have granular control over whether or not the dependency check "report" itself is enforced. This would be a more involved PR, so I'd like to do it after the 'quick fix' to comment-summary-in-pr: on-failure.

**Describe the solution you'd like**
Add an optional input parameter for `dependency-review-action` to enable the creation of a [commit status check](https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#create-a-commit-status):
* If no dependency issues are found or `warn-only: true` is set, the commit state will be set to `success`.
* If dependency issues are found and `warn-only: false` (or the default) is set, the commit state will be set to 'failure'.
* The `target_url` of the commit status will point at the workflow run containing the action. This allows users to easily navigate to a full dependency report.
* Commit status checks should be used instead of [check runs](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#create-a-check-run) because check runs only offer support for GitHub Applications. Creating a check run from a workflow action can result in the check run being randomly attached to a different workflow since all workflows get grouped together into the same check suite.

**Describe alternatives you've considered**
* Leaving the required workflow in `Evaluate` mode and setting `warn-only: false`. This would make it impossible to enforce a requirement to address vulnerable dependencies without also enforcing the workflow running. At an enterprise-level it would be a negative user experience if pull requests were blocked due to a workflow failure (edge case).
* Set `warn-only: true` on shared required workflow. Setting this value wouldn't allow repository administrators to opt-in to enforcing that dependency review problems be addressed in pull requests

**Additional context**
This issue was created to separate out discussion from https://github.com/actions/dependency-review-action/issues/817#issuecomment-2307721535

PR #827, opened 2024-09-06
fix: add summary comment on failure when warn-only: true

### LINKED ISSUE #817: Add comment when warn-only: true and comment-summary-in-pr: on-failure
I'm looking to deploy dependency-review-action using a [required workflow](https://docs.github.com/en/enterprise-cloud@latest/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/available-rules-for-rulesets#require-workflows-to-pass-before-merging) to ensure that all developers are aware of dependency issues when pull requests are opened. 

Since the required workflow rule will be applied to every repository:
1. We don't want the workflow to fail with an error when a dependency issue is found, only report on them so developers are informed
2. We don't want a dependency review summary comment added to every pull request, only ones with issues

The action input `warn-only: true` disables errors. The action input `comment-summary-in-pr: on-failure` avoids adding a comment on every pull request. However, they are used at the same time, no comment summary is added. This is because the comment summary is only added if the entire action will fail with an exit code (`core.setFailure`) not if a dependency review "rule" fails.

Would you be open to having me submit a pull request for the following?
* Modify `comment-summary-in-pr: on-failure` so that a comment is added if any dependency review failure occurs, even if `warn-only: true` is set.

In addition, if you're open to additional changes, I'd like to open a second pull request to have dependency review optionally create a [pull request status check](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#create-a-check-run) that is independent of the workflow. This would allow the dependency review action to work similarly to CodeQL - the success of the action itself would be (optionally) independent from the "dependency failures", allowing developers to enforce that dependency review is "run" but still have granular control over whether or not the dependency check "report" itself is enforced. This would be a more involved PR, so I'd like to do it after the 'quick fix' to `comment-summary-in-pr: on-failure`.

ebickle:
Fixes #817 

#### Summary
When the parameter `warn-only` is set to `true`, allow a pull request comment to be created when `comment-summary-in-pr` is set to `on-failure`.

#### Testing
The action was tested with the following scenarios:
1. New package with vulnerability in pull request, `warn-only: true`, `comment-summary-in-pr: always`
2. New package with vulnerability in pull request, `warn-only: true`, `comment-summary-in-pr: on-failure`
3. New package with vulnerability in pull request, `warn-only: true`, `comment-summary-in-pr: never`
4. No changes to dependencies in pull request, `warn-only: true`, `comment-summary-in-pr: always`
5. No changes to dependencies in pull request, `warn-only: true`, `comment-summary-in-pr: on-failure`
6. No changes to dependencies in pull request, `warn-only: true`, `comment-summary-in-pr: never`

Unit tests have not been added since both `main.ts` and `comment-pr.ts` currently lack unit test files. I looked at adding unit tests for `comment-pr.ts` but the static octokit constructor (`new retryingOctokit`) made this extremely difficult due to `jest.mock` hoisting. To make the file more testable, `retryingOctokit` should be replaced with non-static logic, such as an update to `octokitClient` in `utils.ts` that would have an optional parameter to allow retries. This change would have been much larger and far beyond the scope of this pull request.

#### Implementation decisions
* The fix requires the `config.comment_summary_in_pr === 'on-failure'` test in `comment-pr.ts` to no longer rely on a non-zero exit code. To signal a failure, a new parameter was added to the `commentPr` function.
* A `failureCount: number` was used instead of a boolean to simplify the logic - it's much easier to incrementally do a `failureCount +=` than to try to do boolean logic on the same line as function calls. In addition, `failureCount` seemed more useful for future action logging.
* To be consistent with the rest of the action's functionality, the logging and `setFailed` within `deny.ts` was moved out into `printDeniedDependencies`.
* The `failureCount` variable in `main.ts` can be used for other future functionality, such as an implementation of #825


elireisman (approved):
Thank for you contributing this change! Also really appreciate the manual testing, this functionality is obtuse to unit test at the moment. That too is being tracked, appreciate you working around it for now 🙇 

I'd be partial to sticking with the bool flag for fails until we have time for a larger refactor, but no need to deal with that now since we're not relying on the returned count itself yet anywhere public facing 👍 

Before landing this change, you'll need to update the PR (NPM dist packaging) but otherwise LGTM

ebickle:
I've temporarily converted this back to a draft PR - I've updated to code to switch to using a boolean instead of a failure count. Will mark ready for review once I've run through the manual tests.

ebickle:
@elireisman I've updated the PR to use booleans instead of a numeric "issue count" variable. I also noticed that `core.group` returns a promise and the inner lambda is async, so as a matter of correctness I've swapped over to using promises for the return values to avoid the potential for race conditions.

Should be ready for review again now!

elireisman:
👋 thanks again @ebickle - our team FR will have a look ASAP so we can get this shipped, appreciate it!

Ahmed3lmallah (approved):
👋 @ebickle, thanks again for your contribution. This LGTM. I will merge the PR and make a new release shortly after.

Issue #786, opened 2024-06-25
Job Summary Size Limitation aborts the job [BUG] 


Shweta4398:
**Describe the bug**
It gives me the below error of the size .

**To Reproduce**

Warning: Unable to write summary to pull-request. Make sure you are giving this workflow the permission 'pull-requests: write'.
Error: $GITHUB_STEP_SUMMARY upload aborted, supports content up to a size of 1024k, got 2906k. For more information see: https://docs.github.com/actions/using-workflows/workflow-commands-for-github-actions#adding-a-markdown-summary

My workflow uses upgraded version that is :-
      - name: 'Dependency Review'
        id: dependency-review
        uses: actions/dependency-review-action@72eb03d02c7872a771aacd928f3123ac62ad6d3a
        with:
          base-ref: "${{ github.event.pull_request.base.sha || 'master' }}"
          head-ref: "${{ github.event.pull_request.head.sha || github.sha }}"
          deny-licenses: LGPL-2.0, BSD-2-Clause, LGPL-3.0, LGPL-3.0+
           # the comment-summary-in-pr must be present, otherwise you won't get the output 
          comment-summary-in-pr: always
        continue-on-error: true

osterman:
We just started getting this as well.
https://github.com/cloudposse/atmos/actions/runs/9664474935

Shweta4398:
Hello @osterman ,

Can we have some fix for this?? It is breaking the workflow .


Best regards,
Shweta Singh.

Shweta4398:
@juxtin Can you help here ??

jesus-linares:
any update on this? thanks

kimrtaylor:
This is affecting us too, even after updating to v4.3.3+ 

It's causing confusion within our team because it appears as a job failure. So until this is fixed we don't think we can use this GH action

jonjanego:
Hi folks, Jon from GitHub here. Thanks for filing this issue originally @Shweta4398 and others for commenting on it.

What's happening here is that the action summary buffer is exceeding its maximum size and we are not truncating it gracefully, which then triggers the error that you're getting in the logs from `actions/core`. Our team is starting to take a look into the problem.

A root cause of this would be on PRs with a massive amount of changes, such as [the one](https://github.com/cloudposse/atmos/pull/612) where this was found by your team @osterman, in a PR that reorganized substantial parts of the repo. Because this action writes a summary of all the dependency changes to the action job summary, it can get into this situation.

[...]

PR #828, opened 2024-09-13
Do not list changed dependencies in summary

hmaurer:
Closes https://github.com/actions/dependency-review-action/issues/786.

## Problem

Right now we list every dependency that was either added or removed in the Markdown job summary. For pull requests with large number of changes, this goes over the summary size limit, as reported in https://github.com/actions/dependency-review-action/issues/786.

## Changes in this pull request

I removed the list of changed dependencies from the job summary. They can still be found in the logs. My rationale for this change is that the list of changed dependencies is likely not what users most care about in the summary: the list of vulnerable dependencies, license infringements and such are what's key. The way we were currently listing the dependency changes in the summary was not of much value either way as we were not differentiating between added dependencies and removed dependencies, we were listing all of them (whether added or removed).

This is not the most elegant way to address this issue, but it is the most straightforward.


src/comment-pr.ts
```diff
@@ -6,6 +6,7 @@ import {RequestError} from '@octokit/request-error'
 import {ConfigurationOptions} from './schemas'
 
 export const MAX_COMMENT_LENGTH = 65536
+export const MAX_GITHUB_STEP_SUMMARY_LENGTH = 1024 //1kb
```
jonjanego, line 9: the summary max is 1048576 bytes, not 1024.  it is 1MiB.  although i don't see where this constant is being used?

scripts/create_summary.ts
```diff
@@ -143,7 +143,7 @@ async function createSummary(
     ...licenseIssues.unlicensed
   ]
 
-  summary.addScannedDependencies(allChanges)
+  summary.addScannedFiles(allChanges)
```
jonjanego, line 146: this still doesn't totally protect against the overflow scenario, but yes, much less likely that there would be 1MB worth of filechanges

jonjanego (commented):
do you have a test run of this i could see?

scripts/create_summary.ts
hmaurer: @jonjanego that's right, we could also do a check against the overflow scenario (1048576 bytes) and trim the summary if so. I omitted it because, as you say, without listing the individual dependency changes it's exceedingly unlikely the summary would go over 1MB.

jonjanego: if it is not _too much_ effort, it may be good to do that just to assure ourselves of not hitting this failure again.  I started to take a crack at it, but backed away due to time constraints... but maybe there is a simpler way than how i was looking at it.

elireisman (approved):

scripts/create_summary.ts
elireisman: Updated the PR to handle this too ✅ ready for review 👀

Issue #854, opened 2024-12-06
[BUG] action does fail with valid licenses like Unicode-3.0


ailox:
**Describe the bug**
This action fails when adding a license like [Unicode-3.0](https://spdx.org/licenses/Unicode-3.0.html) into the `allowed_licenses` section.

**To Reproduce**
1. Add the dependency-review-action to a project
2. allow-licenses: "Unicode-3.0"
3. Trigger
4. See error: "Error: Invalid license(s) in allow-licenses: Unicode-3.0"

**Expected behavior**
License is allowed and can be used.

**Action version**
4.5.0 (latest as of today)

ailox:
I used this primitive test to validate the solution:

```ts
import {expect, test} from '@jest/globals'
import * as spdx from '../src/spdx'

test('Unicode', async () => {
  expect(spdx.isValid('Unicode-3.0')).toEqual(true)
})
```

PR #855, opened 2024-12-06
Update transitive dependency spdx-license-ids

ailox:
New licenses were added 7 months ago in https://github.com/jslicense/spdx-license-ids/commit/6a6cd1db9e6526506b6bdd83b592e1512420f1e1, but this logic is not used yet.

This PR fixes https://github.com/actions/dependency-review-action/issues/854 as it updates the transient dependency, so that it is possible to use the Licenses.


ailox:
If helpful, I can also add a test that uses one of the newer licenses, so the resolution can be validated

brrygrdn:
👋🏻 Thanks for this contribution @ailox, this looks good to me. Would you mind running `npm run package` and checking in the dist changes on your branch? All the other checks are 👍🏻. 

I don't think it's worth adding a unit specifically for the new licenses, but thanks for bringing it up!

ailox:
I ran the command you supplied. I was struggling a little since the project had to be `npm build` first

graysonguarino:
Have there been any updates on this PR? We are seeing the same issue and are curious when a fix will be merged.

brrygrdn (approved):

Issue #887, opened 2025-01-27
[BUG] Action fails when removing a dependency version specified in deny_packages


ellenfieldn:
**Describe the bug**
If a repository already has a dependency that is specified as part of `deny_packages`, the dependency review action will block any commit that attempts to remove that dependency.

**To Reproduce**
1. Create a dotnet project using FluentAssertions@8.0.1. Ensure that this change is merged into the default branch.
2. Introduce a GitHub Action workflow that will use the dependency review action and add `deny-packages: pkg:nuget/FluentAssertions@8.0.0, pkg:nuget/FluentAssertions@8.0.1`. 
3. Submit a Pull Request that changes the version of FluentAssertions that is being used to version 7.1.0 (a version that is not explicitly denied.)
4. The dependency review action will return an error status and will report that the version in the previous commit is a denied dependency.

**Expected behavior**
Dependency review action should not block a commit that **removes** a dependency that is not allowed.

**Screenshots**
<details><summary>Job Output and PR Comment</summary>
<p>

Screenshot showing the job failing along with relevant output
![Image](https://github.com/user-attachments/assets/5645aefb-ff72-41f1-b287-2e3aac9e221f)

Screenshot showing the PR comment that is generated as a result of the job
![Image](https://github.com/user-attachments/assets/fc861eea-a04b-4fe7-a2e6-a6d7d15251f0)

</p>
</details> 

**Action version**
v4.5.0 - `uses: actions/dependency-review-action@v4.5.0`

**Examples**
I don't have a public example, but here's a snippet with the workflow
<details><summary>Dependency Review Action</summary>
<p>

``` yaml
name: 'Copy of Workflow'
on: [pull_request]

permissions:
  contents: read
  packages: read
  pull-requests: write

jobs:
  dependency-review:
    runs-on: ubuntu-latest
    steps:
      - name: 'Checkout Repository'
        uses: actions/checkout@v4
      - name: Dependency Review
        uses: actions/dependency-review-action@v4.5.0
        with:
          fail-on-severity: moderate
          deny-licenses: GPL-3.0
          deny-packages: pkg:nuget/FluentAssertions@8.0.0, pkg:nuget/FluentAssertions@8.0.1
          fail-on-scopes: development, runtime
          comment-summary-in-pr: always
```
</p>
</details> 

If you have encountered a problem with a specific package (e.g. issue with license or attributions data) please share details about the package, as well as a link to the manifest where it's being referenced.

[...]

PR #888, opened 2025-01-27
Allow deny package removal

ellenfieldn:
## Summary
There was a bug where if a repository already has a dependency specified as part of deny_packages, the dependency review action will block any commit that attempts to remove that dependency. See #887 

In `deny.ts`, there was no logic around whether the dependency was being added or removed as part of the change. By comparison, licenses are only rejected when dependencies are added or updated (which I believe shows up as two distinct changes -- an add and a remove).

This PR updates the behavior so that pre-existing dependencies not allowed as part of `deny_packages` can be removed. This works for each of the key scenarios:
- Deny a package by name
- Deny a package by name and version
- Deny a package by namespace


AshelyTC:
👋 @ellenfieldn thank you for contributing! There seems to be an error with the generated `/dist` files — it's failing in one of our CI checks. Could you try rebuilding the changes by running `npm run build && npm run package`?

ellenfieldn:
> 👋 @ellenfieldn thank you for contributing! There seems to be an error with the generated `/dist` files — it's failing in one of our CI checks. Could you try rebuilding the changes by running `npm run build && npm run package`?

Woops! Sorry about that, I think i crossed some streams when I merged from main. 

Ran the commands and also enabled and ran the action in my fork just to be sure.

AshelyTC (approved):

PR #893, opened 2025-02-07
Fix typos

omahs:
Fix typos


ahpook (approved):
Good catches, thanks for the improvements!

Issue #897, opened 2025-02-20
[BUG] validity could not be determined for `(MIT OR Apache-2.0) AND Unicode-3.0`


altendky:
**Describe the bug**

At a glance, `(MIT OR Apache-2.0) AND Unicode-3.0` looks to me like a valid SPDX license but dependency-review-action is having trouble with it.

https://github.com/Chia-Network/clvm_tools_rs/actions/runs/13433337446/job/37529914688?pr=88
```
Run actions/dependency-review-action@v4
Vulnerabilities
Licenses
  Warning: 
  The validity of the licenses of the dependencies below could not be determined. Ensure that they are valid SPDX licenses:
  Cargo.lock » unicode-ident@1.0.17 – License: (MIT OR Apache-2.0) AND Unicode-3.0
  wasm/Cargo.lock » unicode-ident@1.0.17 – License: (MIT OR Apache-2.0) AND Unicode-3.0
  Error: Dependency review could not detect the validity of all licenses.
```

It appears that the package is also setting this correctly.

https://github.com/dtolnay/unicode-ident/blob/1.0.17/Cargo.toml#L10
```toml
license = "(MIT OR Apache-2.0) AND Unicode-3.0"
```

**To Reproduce**

**Expected behavior**
The license is able to be processed and is considered valid.

**Screenshots**

![Image](https://github.com/user-attachments/assets/c2cab499-5a14-410d-80e0-b6a4c05251bd)

**Action version**
What version of the action are you using in your workflow? 

We are using `@v4` and this ran two hours ago so it ought to be the latest v4.5.0 from November.

_Note: if you're not running the [latest release](https://github.com/actions/dependency-review-action/releases/latest) please try that first!_

**Examples**
https://github.com/Chia-Network/clvm_tools_rs/actions/runs/13433337446/job/37529914688?pr=88
https://github.com/dtolnay/unicode-ident/blob/1.0.17/Cargo.toml#L10

**Additional context**
n/a

altendky:
I don't think they are entirely unsupported, but I've certainly run into various issues.

https://github.com/actions/dependency-review-action/pull/719
> The `MIT` license satisfies the expression `MIT OR GPL-2.0`, but it does not satisfy `MIT AND GPL-2.0`.

jtomkiew-mng:
[...]

Issue #907, opened 2025-03-17
[BUG] MIT License identified as invalid starting on version 4.3.4


mathiasquatorze:
**Describe the bug**
Starting on version v4.3.4, the MIT license began displaying the following error: 
```  
  The following dependencies have incompatible licenses:
  .github/workflows/dependabot-dependency-review.yml » actions/dependency-review-action@4.*.* – License: MIT
  Error: Dependency review detected incompatible licenses.
```
The action is configured as: 
```
  dependency-review:
    runs-on: ubuntu-latest
    steps:
      - name: "Checkout Repository"
        uses: actions/checkout@v4
      - name: "Dependency Review"
        uses: actions/dependency-review-action@v4
        with:
          config-file: './.github/linters/dependabot-config.yml'
```
And the config file: 
```
fail_on_severity: 'critical'
comment-summary-in-pr: false
allow-licenses:
  - '0BSD'
  - 'Apache-2.0'
  - 'Apache-2.0 AND BSD-3-Clause AND Python-2.0'
  - 'Apache-2.0 AND ISC'
  - 'Apache-2.0 OR (Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause)'
  - 'Apache-2.0 OR MPL-1.1 OR (Apache-2.0 AND MPL-1.1)'
  - 'Apache-2.0 OR MS-PL OR (Apache-2.0 AND MS-PL)'
  - 'Apache-2.0 OR Unlicense OR (Apache-2.0 AND Unlicense)'
  - 'BlueOak-1.0.0'
  - 'BSD-2-Clause'
  - 'BSD-2-Clause AND BSD-3-Clause'
  - 'BSD-2-Clause AND ISC'
  - 'BSD-2-Clause OR (MIT OR Apache-2.0)'
  - 'BSD-3-Clause'
  - 'BSD-3-Clause AND MIT'
  - 'BSD-3-Clause AND BSD-3-Clause-Clear'
  - 'CC-BY-4.0'
  - 'ISC'
  - 'ISC AND MIT'
  - 'MIT' # <--- MIT IS HERE
  - 'MIT-0'
  - 'MIT AND MPL-2.0'
  - 'MIT AND PSF-2.0'
  - 'MIT AND Python-2.0'
  - 'MIT AND Zlib'
  - 'MIT OR (CC0-1.0 AND MIT)'
  - 'MIT OR WTFPL'
  - 'MIT OR WTFPL OR (MIT AND WTFPL)'
[...]

PR #915, opened 2025-04-07
fix: Support SPDX expressions with operators in allow/deny license lists

jebeaudet:
Previously, SPDX expressions with logical operators (e.g., "MIT OR Apache-2.0") were not evaluated correctly when passed as entries in allow or deny lists because the parameters are inversed when using the `@onebeyond/spdx-license-satisfies` library. Per [the readme](https://github.com/onebeyond/spdx-license-satisfies) : 
<details>
<summary>Library documentation</summary>

## `satisfiesAny(first, second)`

Checks if `first` satisfies any license in `second`.

- `first` must be a string with a license or a [SPDX expression](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/). Disjunctive `OR` operator, conjunctive `AND` operator and exception `WITH` operator are allowed.
- `second` must be an array of licenses. Disjunctive `OR` operator and conjunctive `AND`operator are **not** allowed.

```js
const { satisfiesAny } = require('spdx-license-satisfies');

satisfiesAny('(MIT AND GPL-1.0) OR Apache-2.0', ['Apache-2.0']) // true
satisfiesAny('MIT OR GPL-1.0', ['Apache-2.0']) // false
```
</details>

The allow/deny list need to be the first parameter of the library method invocation which was not the case since https://github.com/actions/dependency-review-action/pull/719.

This commit updates the license evaluation logic to properly interpret each expression in the list using `spdx.satisfiesAny`, ensuring that complex SPDX license expressions are supported.

Adds tests to cover:
- Allow-list expressions that match or fail based on license logic
- Deny-list expressions that match or skip based on license logic

fixes gh-897 gh-907


Ahmed3lmallah (approved):
Great find, thank you for fixing this and for adding tests to properly test the behavior!

jebeaudet:
@Ahmed3lmallah Thanks for the review.

Unfortunately, by talking with colleagues, I learned that libraries can have SPDX expressions like `MIT AND Apache-2.0` ([example](https://github.com/krxecs/dual-licensed-mit-apache)) which will not work with this PR since SPDX operators are forbidden in the `second` parameter.

Given this, I'll close this PR since I don't believe it can be merged as is. 

Having spent way to much time trying to debug this in my organization, I think the crux of the issue here is that https://github.com/actions/dependency-review-action/pull/719 removed the possibility to use SPDX expression with `OR` or `AND` in them and people had them in their `allow-list` since it worked before. This limitation was added in https://github.com/actions/dependency-review-action/releases/tag/v4.3.4, in a patch release, without any warning with no mention of which underlying issue still exist.

PR #926, opened 2025-05-05
Replace OTHER with a LicenseRef

dangoor:
## Purpose

ClearlyDefined uses the string `OTHER` for the declared license when
a human has reviewed `NOASSERTION` text and found it to be a valid
license, but one without an SPDX identifier. `OTHER`, unlike
`NOASSERTION`, is not valid. With this change, when `OTHER` appears
in a license string, we'll replace it with
`LicenseRef-clearlydefined-OTHER`, which _is_ valid and will allow
the expressions to parse.

## Related Issues

N/A


src/spdx.ts
```diff
@@ -47,10 +50,19 @@ export function satisfiesAll(
...
     return false
   }
 }
+
+const replaceOtherRegex = /(?<![\w-])OTHER(?![\w-])/
+
+// adjusts license expressions to not include the invalid `OTHER`
+// which ClearlyDefined adds to license strings
+export function removeInvalidSPDX(spdxExpr: string): string {
+  return spdxExpr.replace(replaceOtherRegex, 'LicenseRef-clearlydefined-OTHER')
```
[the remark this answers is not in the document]
dangoor: This is a good point. I'll change this.

src/spdx.ts
```diff
@@ -14,6 +14,7 @@ export function satisfies(
   candidateExpr: string,
   constraintExpr: string
 ): boolean {
+  candidateExpr = removeInvalidSPDX(candidateExpr)
```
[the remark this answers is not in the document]
dangoor: This is _probably_ unnecessary since the constraint expressions are written by the users and `OTHER` is invalid already. Users should write `LicenseRef-clearlydefined-OTHER`. That does bring to mind an issue with the approach in this PR: when an invalid license message is shown, it would still say `OTHER` instead of `LicenseRef-clearlydefined-OTHER`, so it wouldn't give users a clue about what they'd need to add to their license list. (Generally speaking, they shouldn't be adding OTHER in any form to their allowed license list because they can't know what the terms of that license are!)

__tests__/spdx.test.ts
```diff
@@ -73,6 +83,7 @@ describe('satisfiesAll', () => {
       licenses: ['MIT'],
       expected: true
     },
+    // TODO(dangoor): this does not seem correct to me: the only license is Apache-2.0 and it's on the list
```
Ahmed3lmallah, line 86: Given that this tests `satisfiesAll`, I'm assuming that the candidate should satisfy all entries in the `licenses` list, which seems to be correctly not the case here.

__tests__/spdx.test.ts
```diff
@@ -130,6 +141,16 @@ describe('satisfiesAll', () => {
       candidate: 'MIT OR ISC',
       licenses: ['MiT'],
```
Ahmed3lmallah, line 142: Not related to this PR, but here is another case where I would have expected the result to be `true` but it's likely not because it's case sensitive. Does the matching here need to be case sensitive?

src/spdx.ts
```diff
@@ -47,10 +50,19 @@ export function satisfiesAll(
...
   } catch (_) {
     return false
   }
 }
+
+const replaceOtherRegex = /(?<![\w-])OTHER(?![\w-])/
+
+// adjusts license expressions to not include the invalid `OTHER`
+// which ClearlyDefined adds to license strings
+export function removeInvalidSPDX(spdxExpr: string): string {
```
Ahmed3lmallah, line 66: The method name initially threw me off. I thought we are removing an entry from a list, but after looking at the data types I realized I was wrong. Perhaps something like `cleanInvalidSPDX` would be more clear?

Ahmed3lmallah (dismissed):
LGTM besides minor comments. The dist needs to be regenerated too.

dangoor:
I'm going to need to revise this a little based on #927.

__tests__/spdx.test.ts
dangoor: Hmm, so the official spec says that [license identifiers are case-insensitive](https://spdx.github.io/spdx-spec/v3.0.1/annexes/spdx-license-expressions/#case-sensitivity).

This, then, is not actually the _correct_ behavior as per the spec, but is the behavior expected of the library we're using. Definitely out of scope for this PR.

src/spdx.ts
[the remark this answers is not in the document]
dangoor: I spoke with @ahpook about this particular UX nit and he wasn't worried about it in the context of all of the good this change does in general. (Plus, from a compliance perspective, people really should not be adding `LicenseRef-clearlydefined-OTHER` to their allow lists because you can't be sure what the terms you're agreeing to are.)

dangoor:
Oof, I didn't realize I had been working from a fork and not a branch. I pushed a change that is rebased on #927.

dangoor:
This is ready for review again, now that #927 has landed.

juxtin (approved):

PR #927, opened 2025-05-05
Handle complex licenses (e.g. X AND Y)

dangoor:
## Purpose

There are many packages that are dual-licensed, offering a choice of licenses (e.g. `MIT OR Apache-2.0`). There are some that include code from multiple sources and require multiple licenses (e.g. `MIT AND Apache-2.0`). There are also complex combinations that can exist for a variety of reasons, such as `MIT AND (Apache-2.0 OR BSD-3-Clause)`.

The most straightforward approach to handle these is to have an allow list. As long as the licenses on the allow list can satisfy the license expression of the package in question, it should pass.

To implement this, I the newest release of spdx-satisfies which changed the interface to be exactly as described `satisfies(license, allowList)` (see https://github.com/jslicense/spdx-satisfies.js/pull/17).

## Related Issues

Fixes https://github.com/actions/dependency-review-action/issues/263


dangoor:
👋  Hi @jebeaudet! I'm new to maintaining dependency-review-action but have been working on GitHub's internal license compliance tooling for more than 2 years, so I've been staring at license-related issues for longer than I'd prefer to admit 😄  

From my experience, there are _a lot_ of packages with licenses that come from ClearlyDefined that are various combinations of other licenses for a variety of reasons. Requiring all of those combinations to appear in an allow list would cause the allow list to grow dramatically and be unwieldy. Especially if you're using an org-wide allow list. Additionally, comparing two SPDX expressions [can be hard to reason about](https://github.com/jslicense/spdx-satisfies.js/pull/17).

When it comes to answering the question "are we okay using package X?" having an allow list of license IDs and seeing if that list satisfies the license expression of the package works quite well and is very easy to reason about. In general, I suspect most companies would put `MIT` and `Apache-2.0` on that list. Licenses may have different actions you need to take when you distribute the package, but the basic question of "are we okay with this?" should be answerable by most people using just a license list.

So, I took things in a different direction from your recent change, and this turned out to be pretty straightforward given the 6.0.0 release of spdx-satisfies that came out earlier this year.

jebeaudet:
Hi @dangoor!

While I'm happy to finally see some tractions on license checks in GHAS, I don't think we're heading into the right direction with this PR.

I cannot confirm 100% since your PR does not build, but the test named `it handles allowed complex licenses` is wrong. A dual license project with `MIT AND Apache-2.0` cannot be accepted with an allow-list consisting of the 2 separate licenses individually (ie. `['MIT', 'Apache-2.0']`). I've already explained the reasoning [here](https://github.com/actions/dependency-review-action/pull/916#issuecomment-2816889099), backed my the legal team of my company. I encourage you to use GitHub's legal team to confirm on your end.

If there's no way for a company to define the list of dual/triple license allowed with valid SPDX identifiers, it renders the tool worthless from a compliance standpoint. 

I can understand the pain of having a long allow-list but from the research I've done, it's a necessary evil if you want to properly allow/deny the licenses you use, depending on the distribution type of the software you build.

The list could stay relatively short if clearlydefined and/or scancode would fix duplicates in their detection, having duplicates in the SPDX expression is not optimal, for example [aspectjweaver](https://clearlydefined.io/definitions/sourcearchive/mavencentral/org.aspectj/aspectjweaver/1.9.24) who has a whopping `EPL-2.0 AND BSD-3-Clause AND Apache-1.1 AND EPL-2.0 AND EPL-1.0) AND Apache-1.1 AND Apache-2.0 AND BSD-2-Clause AND Apache-2.0 AND Apache-1.1 AND BSD-3-Clause` which could be summarized, removing duplicates, with `EPL-1.0 AND EPL-2.0 AND BSD-2-Clause AND BSD-3-Clause AND Apache-1.1 AND Apache-2.0`. On top of removing duplicate, sorting the `AND` licenses would keep the allow-list at an optimal length.

I'll link one more issue that makes clearlydefined, and therefore GitHub, detect a lot of libraries with `LicenseRef-scancode-unknown-license-reference` : https://github.com/clearlydefined/crawler/issues/636. With this PR, we would have to allow `LicenseRef-scancode-unknown-license-reference` which would be dangerous to allow a project with only this unknown license, instead of working around the issue in your allow-list with SPDX identifiers like `MIT AND LicenseRef-scancode-unknown-license-reference`.

Thanks again for working on this issue!

dangoor:
@jebeaudet Thanks for the reply!

In that comment from your PR, one of the statements near the end that more-or-less sums it up is: "So the combined license might have more obligations than either one alone."

This is definitely true. However, if an organization is fine with the terms of both licenses, then the fact that both licenses are present _simultaneously_ doesn't, in general (see below), change their approval of the licenses. Concretely, MIT and Apache-2.0 are both considered "permissive" licenses and if a package includes both sets of terms, it's still fine. Yes, there are more terms to comply with, but the organization has no problem doing so.

Imagine if, in your whole system, you have one package that's MIT-licensed and another that's Apache-2.0-licensed. When distributing that system, you still need to adhere to the terms of both of those licenses. Allowing `MIT` and `Apache-2.0` in your system still fundamentally requires supporting the combination of terms upon distribution because in many ecosystems you'll come across both of those licenses.

Because I've been looking at this problem for a long time, I will acknowledge that there _are_ some cases of incompatible licenses out there, but they're relatively rare. Among licenses that have gained traction, the open source community has tended to fix compatibility issues by creating new versions of the licenses. The [Python license](https://en.wikipedia.org/wiki/Python_License) had a GPL compatibility issue at one time, as did the [Mozilla Public License](https://en.wikipedia.org/wiki/Mozilla_Public_License), and both of those were updated to fix the issues.

Some tools out there even hide the license list from users, instead offering groupings like "Permissive Licenses" and "Copyleft".

Regarding `LicenseRef-scancode-unknown-license-reference`: That definitely can't go on an allow list (nor could `NOASSERTION` and any variation of that idea). Similarly, I wouldn't put `MIT AND LicenseRef-scancode-unknown-license-reference` on an allow list either, in case another package happened to come up with that combination. I wouldn't put `GPL-2.0` on the allow list, but a GPL-licensed tool used only at build time would be fine. For cases like these, the solution is to acknowledge that the _package's_ license and use has been reviewed. In DRA, you can do that with the `allow-dependencies-licenses` config option.

Also, along those lines, I'll note that our internal (not DRA-based) org-wide allow list does _not_ include `LGPL` specifically because the terms of that license can get complicated depending on how the code is distributed. So we end up either approving the license for particular repos or approving specific packages for specific repos.

What GitHub has been doing internally for a number of years, starting with [licensed](https://github.com/licensee/licensed) and more recently with the tooling I've been working on, looks like what I've described: an allow list of licenses and specific package approvals after they've been reviewed.

Thanks again for contributing here!

dangoor:
I've opened a [pull request for DefinitelyTyped to fix the type errors here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/72702)

jebeaudet:
About the `LicenseRef-scancode-unknown-.*` problem, I agree that it should not be put on the allow-list, I was raising this point trying to get traction on this issue that has been plaguing our pipeline for a long time. It affects some really popular libraries like Jetty, Google errorprone, even [Spring](https://clearlydefined.io/definitions/maven/mavencentral/org.springframework/spring-aop/6.1.18) is in there... Using the `purl` to allow-list can be a workaround, but 1. it's a pain having to add any misdetected license in our allow-list since it blocks the pipeline and requires our legal team to review it and 2. the purl has no concept of versioning, so when you're adding a package there and its license changes, you could technically be in trouble.

About the other point of having SPDX in the allow-list, you have some good points, I've reached out to my legal team to get their opinion, I'll get back to you. I can't help but feel that as a business-grade dependency review tool, not supporting SPDX expressions could be limiting to some clients.

dangoor:
> About the LicenseRef-scancode-unknown-.* problem, I agree that it should not be put on the allow-list, I was raising this point trying to get traction on this issue that has been plaguing our pipeline for a long time.
> Using the purl to allow-list can be a workaround, but 1. it's a pain having to add any misdetected license in our allow-list since it blocks the pipeline and requires our legal team to review it

I agree 💯 %. License data quality is very much on my mind.

> 2. the purl has no concept of versioning range, so when you're adding a package there and its license changes, you could technically be in trouble

I agree that this solution is imperfect. Since most packages never relicense, the risk is relatively low. The approach that I've taken with our internal tool is to track not just the reviewed package but also the licenses it had at the time of review, essentially adding those licenses to a package-specific allow list. So a new license appearing would cause the package to fail again. Maybe an approach like that can ultimately work here.

> 3. purl does not support any king of regex or allowing a groupId so you have to add each one individually.

True. Some kind of wildcard/regex support would be helpful.

> I can't help but feel that as a business-grade dependency review tool, not supporting SPDX expressions could be limiting to some clients.

My opinion is that the user experience tradeoffs (requiring more blocked PRs and updates to the allow list, plus potential confusion around how comparison between more complex SPDX expressions would work) aren't worth it given the points I made above.

brrygrdn (approved):
This change makes sense to me, and I agree with the reasoning on the trade offs discussed above with regard to an allow list permitting `[X, Y]` passing on something that is `X AND Y`.

Issue #801, opened 2024-07-23
[BUG] Listing too many allow-dependencies-licenses makes the summary output unreadable


jtomkiew-mng:
**Describe the bug**
If `allow-dependencies-licenses` input contains a lot of entries, the generated sumary is very difficult to read as it will contain the full list of ignored packages.

**To Reproduce**
1. Using a config file, add ~6000 packages to the `allow-dependencies-licenses`: https://gist.github.com/jtomkiew-mng/5102060143b01ca41ad033d3cbd18f16
2. Run the action
3. Check summary

**Expected behavior**
Limit `Excluded from license check` in the summary to only contain packages that were detected in the execution (i.e. if I have `pkg:nuget/Package.Name` defined in `allow-dependencies-licenses` but this package was not detected in the current workflow, do not list it).

Alternatively, if the above is too much work, simply make `Excluded from license check` as collapsable details (using the `<details>` and `<summary>` tag elements).

**Screenshots**
![image](https://github.com/user-attachments/assets/dc14077b-ad24-4881-9f00-d049da186b72)

**Action version**
v4.3.3

**Examples**
Example config file: https://gist.github.com/jtomkiew-mng/5102060143b01ca41ad033d3cbd18f16
Example project file:
```csproj
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <IsPackable>false</IsPackable>
    <IsTestProject>true</IsTestProject>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.2" />
    <PackageReference Include="System.Net.Security" Version="4.3.2" />
    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
    <PackageReference Include="System.Net.Http" Version="4.3.4" />
    <PackageReference Include="Microsoft.Web.Administration" Version="11.1.0" />
  </ItemGroup>

</Project>
```
Example workflow step:
```yaml
    - name: Dependency Review
      uses: actions/dependency-review-action@v4.3.3
      with:
        license-check: true
        vulnerability-check: true
        warn-only: false
        fail-on-scopes: runtime
        fail-on-severity: high
        config-file: ./dependency-review-config.yaml
```

**Additional context**
None.

jtomkiew-mng:
[...]

PR #954, opened 2025-07-30
#801 Make license lists collapsable

jasperkamerling:
This ensures large allow or deny lists don't create huge comments


brrygrdn:
Merged via #986

PR #986, opened 2025-09-26
Batch some contributions for release

brrygrdn:
This is a manual merge of the following PRs:
- https://github.com/actions/dependency-review-action/pull/954
- https://github.com/actions/dependency-review-action/pull/918

I'm taking this approach to avoid the long feedback loop on asking contributors to rebuild `dist/`

A note to reviewers, the addition of `@actions/artefact` introduces a vulnerability as it depends on an older versions of `@octokit/plugin-paginate-rest@2.21.3`:
- https://github.com/advisories/GHSA-h5c3-5r3r-rr8q

I don't think this is exploitable given the nature of the action and the fact this version is used internally within the artefact package; we only using an upload method without accessing any endpoints that would invoke pagination as far as I can tell.


RomanIakovlev:
The failing DRA check on this PR is caused by the vulnerability you've mentioned.

PR #988, opened 2025-09-26
Bump to 4.8.0

brrygrdn:
The changes shipped very recently as 4.7.4 are reasonable significant, so we should likely be stricter on our use of semver and consider this a minor update rather than patch update.


ahpook (approved):
LGTM! cheers @brrygrdn

PR #1007, opened 2025-10-22
Make handleLargeSummary also update core.summary

gitulisca:
## Why is this change required?

Although we already have a handleLargeSummary function, prior to this change it would return a shorter summary but it would not update the actual core.summary object.

As core.summary was unaltered by that function the [finally](https://github.com/actions/dependency-review-action/blob/main/src/main.ts#L266) block in the `run` function would throw an exception whenever it called core.summary.write() on a core.summary that exceeded the 1024KB maximum Summary size.

summary-size-exceeds-1024k

## Testing

- Updated the code on a fork repository. Built and packaged it.
- Used the updated action on a workflow triggered for a commit containing 900 npm packages which resulted in a summary larger than 1024K. The updated version handled that commit successfully. The original action version didn't.


miguelgonzalez88001-cell:
fail-on-severity: 'critical'
allow-licenses:
  - 'GPL-3.0'
  - 'BSD-3-Clause'
  - 'MIT'

PR #1012, opened 2025-11-05
Generate dist files on main branch

dangoor:
## Purpose

This adapts an approach taken by the Gradle actions in order to
generate the dist files on the main branch rather than having
every contributor need to generate them. (In fact, people will no
longer be able to submit PRs with the dist files updated). This
change is important because the current approach means that
people encounter merge conflicts all the time and will need to
keep regenerating the dist files in order to land their change.

## Related Issues

None


dangoor:
Hmm… I wonder why the check-dist workflow didn't run.

juxtin:
Do we need to forbid updates to the dist directory? Committing the dist files makes it possible to test a branch of this action without releasing it.

dangoor:
> Do we need to forbid updates to the dist directory? Committing the dist files makes it possible to test a branch of this action without releasing it.

Good point! I'll get rid of that check

PR #1013, opened 2025-11-06
Remove bad token reference

dangoor:
## Purpose

The action to build the dist directory referenced a non-existing token.


claire153 (approved):
👍

PR #1017, opened 2025-11-10
GitHub Actions can't push to our protected main

dangoor:
## Purpose

Our main branch is protected, which means that our Actions workflow
cannot push changes directly to main. This removes the non-functional
workflow.


ahpook (approved):
Too bad this didn't work. 🤷

Issue #1046, opened 2026-02-10
allow-dependencies-licenses: allow-list specific license


erikburt:
**Is your feature request related to a problem? Please describe.**

When using the `allow-dependencies-licenses` configuration, it only allows for specifying specific packages and not their particular license. This is an issue if a dependency you've previously allowed, changes their license to something more restrictive.

**Describe the solution you'd like**

Use PURL qualifiers (or subpaths) to allow optional filtering of specific licenses for specific dependencies.

Example:
- `pkg:npm/@foo/bar?license=GPL-1.0-or-later`
  - If this package changes to a different license in my `deny_licenses` field, then it should fail.
- Not specifying license should retain existing behaviour of a "blanket" allow.

**Describe alternatives you've considered**

There are no alternatives.

**Additional context**

N/A

PR #1047, opened 2026-02-10
feat: support package+license specific exclusions

erikburt:
Adds support so `allowedDependenciesLicenses` can be configured for specific licenses. This addresses #1046.

It does this by adding support for a `license` PURL qualifier on the list of allowed dependencies licenses. This maintains backwards compatibility when the `license` qualifier is not included.

See my changes to `example.md` for more information.

### Changes

* Update PURL parser to support `license` qualifier
* Update `licenses.ts` logic to support filtering changes based on `license`
* Update unit tests for purl parser and license logic

#### licenses.ts

1. Moved the filtering logic to it's own function `filterLicenseChange`, rather than it being inlined into `groupChanges`
2. Move the changes filtering to the top-level `getInvalidLicenseChanges` function
    * `groupChanges` function used to prefilter the changes with the exclusion list. It was able to blanket filter here as it didn't matter if a dependency's license had been determined or not. And filtering before the grouping means less licenses to pull.
    * Now, we split the exlcusions list (renamed to `allowedDependenciesLicenses` to match input/config), into 2 separate lists. 1: Those without `?license=...` PURL qualifiers, and those with it.
        * We do this so we can maintain previous behaviour. ie. Any license-agnostic exclusions are applied before we start pulling dependency licenses.
3. Continue `groupChanges` as normal, but apply exclusions that contain `license` qualifiers, after licenses have been determined

### Notes

This does work but if the design is not acceptable, I'm happy to modify it.
- ie. If using a qualifer in the PURL is not ideal or if you'd rather it be another/different input

### DIFF
--- a/__tests__/licenses.test.ts
+++ b/__tests__/licenses.test.ts
@@ -244,7 +244,10 @@ test('it does not filter out changes that are on the exclusions list', async () 
   const changes: Changes = [pipChange, npmChange, rubyChange]
   const licensesConfig = {
     allow: ['BSD-3-Clause'],
-    licenseExclusions: ['pkg:pypi/package-1@1.1.1', 'pkg:npm/reeuhq@1.0.2']
+    allowedDependenciesLicenses: [
+      'pkg:pypi/package-1@1.1.1',
+      'pkg:npm/reeuhq@1.0.2'
+    ]
   }
   const invalidLicenses = await getInvalidLicenseChanges(
     changes,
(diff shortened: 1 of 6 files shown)


johnSmithTech:
Hey @erikburt ,

Came across your GitHub profile while checking recent PR activity — really liked your work 👀

Quick question — do PR reviews ever slow you down?

I’m building Mesrai AI — it connects to GitHub and reviews PRs and changes instantly (bugs, improvements, suggestions) before human review.

You can try it here — free & quick setup (~1 min):

[Mesrai AI](https://app.mesrai.com
)
[Mesrai](https://mesrai.com
)

Would really appreciate your feedback — I’m building this to solve real PR review problems, and your input would be super valuable 🙌

PR #1076, opened 2026-03-23
fix: patched version display for advisories with non-strict semver ranges (e.g. Maven beta versions)

### LINKED ISSUE #1075: [BUG] Patched version shows "N/A" for advisories with non-strict semver version ranges
**Describe the bug**

When show-patched-versions is enabled, the vulnerability summary displays "N/A" as the patched version for advisories whose vulnerable version ranges contain version identifiers that are not strict semver. For example, advisory GHSA-vc5p-v9hr-52mj for org.apache.logging.log4j:log4j-core specifies the range >= 2.0-beta9, < 2.25.3. The version 2.0-beta9 is missing the patch segment required by strict semver (2.0.0-beta9), so semver.validRange() rejects the entire range. This causes the patch version lookup to fail silently, showing "N/A" even though the API returned a valid patched version (2.25.3). This primarily affects Maven packages, which commonly use non-strict version formats like 2.0-beta9 or 2.0-rc1 in their advisory ranges.

**To Reproduce**
1. Create a PR to a pom.xml
2. make sure it includes `log4j` version `2.12.4` (this will trigger advisory [GHSA-vc5p-v9hr-52mj](https://github.com/advisories/GHSA-vc5p-v9hr-52mj))
3. Make sure the dependency review action has the parameter `show-patched-version` set to true
4. See that patched version column shows `N/A`

**Expected behavior**

patched version should show version `2.25.3`

**Screenshots**

Image

**Action version**

v4.9.0

**Additional context**

tspascoal:
When showing patched versions in the vulnerability summary, advisories with version ranges containing non-strict semver components — such as Maven's >= 2.0-beta9, < 2.25.3 — would display "N/A" instead of the correct patched version. This happened because semver.validRange() rejects range bounds like 2.0-beta9 (missing the patch segment, e.g. 2.0.0-beta9), causing the versionInRange function to fail range validation and fall through to its fail-open default of no match. The fix adds a coercion step in fail-open mode that rewrites invalid version components within the range string to their nearest valid semver equivalents (e.g. 2.0-beta9 → 2.0.0) before retrying validation, allowing the patched version lookup to succeed without affecting the fail-closed vulnerability detection path.

Adds test case for maven and another for golang for extra coverage

**Improvements to version range handling:**

* Enhanced the `versionInRange` function in `src/summary.ts` to attempt coercion of invalid version components within the range string for "fail-open" mode, improving support for non-strict semver formats (e.g., converting `2.0-beta9` to `2.0.0-beta9`) and increasing compatibility with real-world advisories.
* Refactored validation logic in `versionInRange` to use a mutable `validRange` variable, preparing for additional range coercion logic.

**Expanded test coverage for vulnerability summary logic:**

* Added a test for handling a Maven package advisory (`GHSA-vc5p-v9hr-52mj`) with a complex version range, ensuring the summary correctly identifies the patched version.
* Added a test for a Go package advisory (`GHSA-r9w3-57w2-gch2`) with a pseudo-version range, verifying correct patched version reporting and API usage.

Closes #1075

### DIFF
--- a/__tests__/summary.test.ts
+++ b/__tests__/summary.test.ts
@@ -696,6 +696,49 @@ test('addChangeVulnerabilitiesToSummary() - handles RestSharp GHSA-4rr6-2v9v-wcp
   })
 })
 
+test('addChangeVulnerabilitiesToSummary() - handles GHSA-vc5p-v9hr-52mj maven range', async () => {
+  const pkg = createTestChange({
+    ecosystem: 'maven',
+    name: 'org.apache.logging.log4j:log4j-core',
+    version: '2.12.4',
+    vulnerabilities: [
+      createTestVulnerability({
+        advisory_ghsa_id: 'GHSA-vc5p-v9hr-52mj',
+        advisory_summary: 'Apache Log4j vulnerable to XXE in JDBC Appender',
+        severity: 'high'
+      })
+    ]
+  })
+
+  // Mock API response from https://api.github.com/advisories/GHSA-vc5p-v9hr-52mj
+  mockOctokitRequest.mockResolvedValueOnce({
+    data: {
+      vulnerabilities: [
+        {
+          package: {
+            ecosystem: 'maven',
+            name: 'org.apache.logging.log4j:log4j-core'
+          },
+          vulnerable_version_range: '>= 2.0-beta9, < 2.25.3',
+          first_patched_version: '2.25.3'
+        }
+      ]
+    }
+  })
+
+  const changes = [pkg]
+  await summary.addChangeVulnerabilitiesToSummary(changes, 'low', true)
+
+  const text = core.summary.stringify()
+
+  // Should show the correct patched version for this advisory range
+  expect(text).toContain('2.25.3')
+  expect(text).not.toContain('N/A')
+  expect(mockOctokitRequest).toHaveBeenCalledWith('GET /advisories/{ghsa_id}', {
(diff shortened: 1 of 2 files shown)


felickz (commented):
The only edge case here would be 


**ex:** Advisory range is `>= 2.0-beta9, < 2.25.3`, patched version is `2.25.3`.

| User's version | Intended range (`>= 2.0.0-beta9`) | Coerced range (`>= 2.0.0`) | Without PR fix (invalid range) |
|---|---|---|---|
| `2.12.4` | match → shows `2.25.3` | match → shows `2.25.3` | N/A |
| `2.0.0` | match → shows `2.25.3` | match → shows `2.25.3` | N/A |
| `2.0.0-beta10` | match → shows `2.25.3` | **no match → N/A** | N/A |

The edge case: a user on `2.0.0-beta10` is **in the vulnerable range** (it's after beta9) but `2.0.0-beta10 < 2.0.0` in semver ordering, so it fails the coerced `>= 2.0.0` check. They'd see "N/A" instead of `2.25.3`.

**But** that user already sees "N/A" today without the fix (the range is completely invalid). So the PR doesn't regress anything; it just doesn't fix this narrow sub-case.  Though the only way you'd get a valid semver prerelease that doesn't match the coerced range is if the ecosystem used semver-style prereleases... but then the range would also be valid semver and no coercion would happen in the first place.

ahpook (approved):
Thanks for the fix @tspascoal

PR #1094, opened 2026-05-08
Resolve security findings

AshelyTC:
- Fixes: https://github.com/github/vuln-mgmt/issues/192264

Resolves _nearly_ all dependabot alerts, except `@typescript-eslint/parser`:
- https://github.com/actions/dependency-review-action/pull/972

It needs more working on since it's a major upgrade and conflicts with peer dependencies.


ahpook (approved):
Nice, thanks @AshelyTC !

PR #1144, opened 2026-08-31
fix: Normalize license ID casing before validation

cadivus:
Some registries report license IDs with non-canonical casing, e.g. pub.dev reports flutter_secure_storage as 'bsd-3-clause'. The SPDX parser is case-sensitive, so these licenses end up in the unresolved bucket and fail the run even when the license is on the allow list.

Map case variants of known SPDX IDs to their canonical form before validating. Unknown tokens are left untouched, and user-provided allow/deny lists are still validated strictly.

Without the fix:  
Screenshot 2026-08-31 at 16 34 04

With the fix:  
Screenshot 2026-08-31 at 16 35 25

The fix with removed BSD-3-Claude (to ensure nothing broke):  
Screenshot 2026-08-31 at 16 36 52


ahpook:
This change looks good but we'll need to resolve the package-lock conflicts - I'm merging a batch of dependabot PRs that will update deps first, then we'll put these new spdx libraries on top

cadivus:
@ahpook Thanks for the review!  Could you  please tell me when to rebase? I wait for you to merge the other dep changes.

ahpook:
Keep an eye on #1150, that's the last one that needs to get merged.

cadivus:
Hi @ahpook,
I just rebased everything.

Zaulao:
Hi everyone! Any updates on this PR? I'm also dealing with several workflows failing due to case mismatches in dependency names, this would really help me out.

